PR #186, opened 2024-08-30
feat(Predictions.PubSub): Publish predictions data periodically & on reset

KaylaBrady:
### Summary

_Ticket:_ [Predictions Scalability: new channel that publishes predictions updates in chunks](https://app.asana.com/0/1205732265579288/1207791938443975/f)

What is this PR for?

This PR replaces `StopPubSub` with a generic `Predictions.PubSub` based on the [PubSub](https://github.com/mbta/dotcom/blob/main/lib/predictions/pub_sub.ex) module in dotcom.

Right now this only includes predictions: A future PR will add the related trips & vehicles as well.

PR #190, opened 2024-09-09
feat(Store.Predictions): Add support for writing + reading predictions

KaylaBrady:
Summary
Ticket: [Predictions Scalability: new channel that publishes predictions updates in chunks](https://app.asana.com/0/1205732265579288/1207791938443975/f)

What is this PR for?

This PR adds only the functionality for reading + writing predictions to the ETS store. The functionality for actually using this ETS table is underway in https://github.com/mbta/mobile_app_backend/pull/186.

### DIFF
--- a/lib/mbta_v3_api/store.ex
+++ b/lib/mbta_v3_api/store.ex
@@ -27,4 +27,9 @@ defmodule MBTAV3API.Store do
   Retrieve all records that match the given filter keys.
   """
   @callback fetch(keyword()) :: [JsonApi.Object.t()]
+
+  @doc """
+  Retrieve all records that match any of the given filter keyword sets
+  """
+  @callback fetch_multi_filter([keyword()]) :: [JsonApi.Object.t()]
 end
--- a/lib/mbta_v3_api/store.ex
+++ b/lib/mbta_v3_api/store.ex
@@ -31,5 +31,5 @@ defmodule MBTAV3API.Store do
   @doc """
   Retrieve all records that match any of the given filter keyword sets
   """
-  @callback fetch_multi_filter([keyword()]) :: [JsonApi.Object.t()]
+  @callback fetch_any([keyword()]) :: [JsonApi.Object.t()]
 end
--- a/lib/mbta_v3_api/store/predictions.ex
+++ b/lib/mbta_v3_api/store/predictions.ex
@@ -5,28 +5,153 @@ defmodule MBTAV3API.Store.Predictions do
 
   Based on https://github.com/mbta/dotcom/blob/main/lib/predictions/store.ex
   """
+  use GenServer
+  require Logger
+  alias MBTAV3API.Prediction
+
   @behaviour MBTAV3API.Store
 
+  @predictions_table_name :predictions_from_streams
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
   @impl true
-  def process_upsert(_event, _objects) do
-    # TODO
+  def init(_) do
+    _table = :ets.new(@predictions_table_name, [:named_table])
+
+    {:ok, %{}}
+  end
+
+  @impl true
+  def fetch(fetch_keys) do
+    match_spec = prediction_match_spec(fetch_keys)
+
+    timed_fetch([{match_spec, [], [:"$1"]}], "fetch_keys=#{inspect(fetch_keys)}")
+  end
+
+  @impl true
+  def fetch_multi_filter(fetch_keys_list) do
+    match_specs =
+      fetch_keys_list
+      |> Enum.map(&prediction_match_spec(&1))
+      |> Enum.map(&{&1, [], [:"$1"]})
+
+    timed_fetch(match_specs, "multi_fetch=true fetch_keys=#{inspect(fetch_keys_list)}")
+  end
+
+  defp prediction_match_spec(fetch_keys) do
+    # https://www.erlang.org/doc/apps/erts/match_spec.html
+    # Match the fields specified in the fetch_keys and return the full prediction
+    # see to_record/1 for the defined order of fields
+    {
+      Keyword.get(fetch_keys, :prediction_id, :_) || :_,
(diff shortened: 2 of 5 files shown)


lib/mbta_v3_api/store/predictions.ex
```diff
@@ -5,28 +5,153 @@ defmodule MBTAV3API.Store.Predictions do
...
+    # https://www.erlang.org/doc/apps/erts/match_spec.html
+    # Match the fields specified in the fetch_keys and return the full prediction
+    # see to_record/1 for the defined order of fields
+    {
+      Keyword.get(fetch_keys, :prediction_id, :_) || :_,
+      Keyword.get(fetch_keys, :route_id, :_) || :_,
+      Keyword.get(fetch_keys, :stop_id, :_) || :_,
+      Keyword.get(fetch_keys, :direction_id, :_) || :_,
+      Keyword.get(fetch_keys, :trip_id, :_) || :_,
+      Keyword.get(fetch_keys, :vehicle_id, :_) || :_,
```
boringcactus, line 55: I feel like if we're using `|| :_` to handle when `fetch_keys` has `vehicle_id: nil` we could just use `Keyword.get/2` so that we don't have to specify the fallback twice.

lib/mbta_v3_api/store/predictions.ex
```diff
@@ -5,28 +5,153 @@ defmodule MBTAV3API.Store.Predictions do
...
+  defp clear_data(keys) do
+    match_pattern = {
+      Keyword.get(keys, :prediction_id, :_) || :_,
+      Keyword.get(keys, :route_id, :_) || :_,
+      Keyword.get(keys, :stop_id, :_) || :_,
+      Keyword.get(keys, :direction_id, :_) || :_,
+      Keyword.get(keys, :trip_id, :_) || :_,
+      Keyword.get(keys, :vehicle_id, :_) || :_,
+      :"$1"
+    }
```
boringcactus, line 132: Should this be calling `prediction_match_spec/1`?

lib/mbta_v3_api/store/predictions.ex
```diff
@@ -5,28 +5,153 @@ defmodule MBTAV3API.Store.Predictions do
...
+  @predictions_table_name :predictions_from_streams
+
+  @spec start_link(Keyword.t()) :: GenServer.on_start()
+  def start_link(_) do
+    GenServer.start_link(__MODULE__, [], name: __MODULE__)
+  end
+
+  @impl true
+  def init(_) do
+    _table = :ets.new(@predictions_table_name, [:named_table])
```
boringcactus, line 23: Did you try making the table name an argument to `start_link/1`/`init/1` and `fetch/1`/`fetch_multi_filter/1`? That feels like it might be nicer than having to mess with the application config to not start the real predictions store, but the fact that `fetch` runs outside the `GenServer` means that the table can't really be anonymous and stored within the `GenServer`, so if it doesn't actually work all that well then the application config solution is definitely acceptable.

lib/mbta_v3_api/store.ex
```diff
@@ -27,4 +27,9 @@ defmodule MBTAV3API.Store do
   Retrieve all records that match the given filter keys.
   """
   @callback fetch(keyword()) :: [JsonApi.Object.t()]
+
+  @doc """
+  Retrieve all records that match any of the given filter keyword sets
+  """
+  @callback fetch_multi_filter([keyword()]) :: [JsonApi.Object.t()]
```
boringcactus, line 34: I think `fetch_any/1` might be a better name than `fetch_multi_filter/1`. Technically, it looks like it'd be possible to just write an ETS query with `or` in `fetch/1` and drop this function entirely, but that may be more work to implement for no particular benefit.

boringcactus (commented):
Sorry it took me so long to get this reviewed.

lib/mbta_v3_api/store.ex
KaylaBrady: 👍 will rename! I initially intended to have a single `fetch` with pattern matching to determine the behavior depending on if the param was a list of keywords or a keyword, but there isn't an `is_keyword` guard. It ended up seeming a bit clearer to me to just have a distinctly named function

lib/mbta_v3_api/store/predictions.ex
KaylaBrady: I did start down that approach but found it clunkier to have to drill down the table name in more places. Testing the `process_` functions would mean also having to take the GenServer name param - ultimately it seemed to mean more points of failure in trying to write a test that works.

lib/mbta_v3_api/store.ex
boringcactus: Ah, there's no `is_keyword` guard, but you could `when is_tuple(hd(keyword_or_list))` if you wanted a similar effect - depending on whether an empty list should match everything or nothing, that may not quite be enough, though.

lib/mbta_v3_api/store/predictions.ex
boringcactus: `Keyword.get/3` will only use its fallback if the requested key is missing from the keyword list; if the key is present but `nil` then it will return `nil`. A hypothetical use case in which I can imagine this mattering:
```elixir
stop_id = Map.fetch!(params, "stop_id")
direction_id =
    if direction_id = Map.get(params, "direction_id") then
        String.to_integer(direction_id)
    end
MBTAV3API.Store.Predictions.fetch(stop_id: stop_id, direction_id: direction_id)
```
If there is no `direction_id` in `params`, then I think this will filter to predictions with a `direction_id` of `nil`. Using `Keyword.get(fetch_keys, :stop_id) || :_` would mean that `stop_id` being `nil` and `stop_id` being absent would be treated the same way, which I think is less likely to cause really confusing bugs in six months.

KaylaBrady: Ah thanks, should have read your initial suggestion more carefully.

boringcactus (approved):

lib/mbta_v3_api/store.ex
KaylaBrady: Ooh there is a `Keyword.keyword?/1` function which isn't quite as good as a guard but seems a bit more readable. I ended up rolling some of this change forward into the next PR

PR #197, opened 2024-09-19
feat(Predictions.PubSub): publish related trip data

KaylaBrady:
### Summary

What is this PR for?

This PR publishes trips associated with predictions by stop. Associated vehicles to follow in a separate PR

### DIFF
--- a/lib/mbta_v3_api/store.ex
+++ b/lib/mbta_v3_api/store.ex
@@ -27,7 +27,18 @@ defmodule MBTAV3API.Store do
   @doc """
   Retrieve all records that match the given filter keys.
   When given a single keyword list, all keys in the list must match.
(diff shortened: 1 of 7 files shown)


KaylaBrady:
Converted to draft - changing the initial message on join to not be partitioned by stop so it is easier to manage in the frontend.

boringcactus (approved):
Nice!

KaylaBrady:
@boringcactus sorry, made one more adjustment to the payload structure on subscribe to include predictions by stop, but trips (and eventually vehicles) still only by ID. 

By keeping predictions by stop ID, it we can just fully replace the entry for that stop on the frontend when a new update comes in.

boringcactus (approved):

PR #198, opened 2024-09-20
feat(Predictions.PubSub): Include related vehicles

KaylaBrady:
### Summary

What is this PR for?

This PR adds a new Vehicle store that is updated by a single stream for all vehicles. 
Those vehicles are fetched as part of fetching predictions for parity with the existing predictions channel.

### DIFF
--- a/lib/mbta_v3_api/store/vehicles.ex
+++ b/lib/mbta_v3_api/store/vehicles.ex
@@ -1,0 +1,199 @@
+defmodule MBTAV3API.Store.Vehicles do
+  use GenServer
+  require Logger
+  alias MBTAV3API.Store.Vehicles
+
+  alias MBTAV3API.Vehicle
+
+  @behaviour MBTAV3API.Store
+
+  def start_link(opts) do
+    Application.get_env(:mobile_app_backend, MBTAV3API.Store.Vehicles, Vehicles.Impl).start_link(
+      opts
+    )
+  end
+
+  @impl true
+  def init(opts) do
+    Application.get_env(:mobile_app_backend, MBTAV3API.Store.Vehicles, Vehicles.Impl).init(opts)
+  end
+
+  @impl true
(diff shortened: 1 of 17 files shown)


lib/mbta_v3_api/store/vehicles.ex
```diff
@@ -0,0 +1,198 @@
+defmodule MBTAV3API.Store.Vehicles do
```
EmmaSimon, line 1: This looks all identical to `Store.Predictions` except for the Store class and the Impl, would it be at all straightforward to make something more generic that could just take those two params and generate the module? Or is it possible/likely that other ones will have different implementations?

EmmaSimon: Er, well, I guess alerts are the only other realtime stream, do we plan to store those in ETS too, or just these two? If it's just going to be two then I'm more fine with the repetition, but three is pushing it.

KaylaBrady: Yeah that's fair! I don't have plans to convert alerts as part of this ticket (but there will be a bunch of follow-up tasks, like converting the vehicles channel to use the new vehicles table & predictions by trip to use the new trips table). 

I did a time boxed pass of making this more generic in a few ways:
* macro to hide the boilerplate calling the correct module for testing
* common timed_fetch function

Predictions & Vehicles are still mostly similar, but Predictions has the extra logic for pulling in other related records. I think any further refactoring to make these common would be worth holding off on.

KaylaBrady:
Where the vehicle stream is subscribed to needs to move - will push that in a bit!

lib/mbta_v3_api/store/vehicles.ex
EmmaSimon: That makes sense, this looks great, thank you!

EmmaSimon (approved):

PR #199, opened 2024-09-20
feat(MBTAV3API): Use ets cache instead of disk cache

KaylaBrady:
### Summary

What is this PR for?

In load testing the new channel, we found `Req` requests that were way longer than the corresponding V3 API request. We tested switching to an in-memory ETS cache (per @paulswartz rec) and found it performed better under load (test results in [Scalability notes](https://www.notion.so/mbta-downtown-crossing/Nearby-Transit-Scalability-Notes-104f5d8d11ea80829652c8f4a0aba748)). 

PR based on 
https://github.com/mbta/screens/pull/2055

Ran locally and confirmed received 304s from the v3 API 
```
5f13054c9cd7 2024-09-20T15:10:43.982 [info] node=api@ip-10-2-30-237 request_id=F_b8axS6PLPTjZMAwL9C api_key=[redacted] api_version=2021-01-09 concurrent=4 state=set duration=0.292 status=304 params={"fields":{"route":"type,color,direction_names,direction_destinations,long_name,short_name,sort_order,text_color"},"filter":{"stop":"place-portr,70066,FR-0034,70065,FR-0034-01,FR-0034-02"}} path=/routes method=GET format=json-api controller=ApiWeb.RouteController action=index
```


config/config.exs
```diff
@@ -15,6 +15,10 @@ config :mobile_app_backend, MobileAppBackend.AppCheck,
 
 config :mobile_app_backend, predictions_broadcast_interval_ms: 10_000
 
+config :mobile_app_backend, MBTAV3API.ResponseCache,
+  gc_interval: :timer.hours(1),
+  allocated_memory: 250_000_000
```
KaylaBrady, line 20: Pulled these values from screens as a starting point

boringcactus (approved):
👍

paulswartz (commented):
I'm not 💯 sure it would be cleaner, but you could also implement this as a [Req plugin](https://hexdocs.pm/req/Req.Request.html#module-writing-plugins).

boringcactus:
If we implemented it as a Req plugin, it'd run inside of `timed_get/3`, which would increase the gap between the request duration logged by the V3 API and the request duration logged by `timed_get/3`/`log_response/4`. I'm not sure how much that matters, though, and we might want the same logic for caching Algolia requests.

paulswartz:
> If we implemented it as a Req plugin, it'd run inside of `timed_get/3`, which would increase the gap between the request duration logged by the V3 API and the request duration logged by `timed_get/3`/`log_response/4`. I'm not sure how much that matters, though, and we might want the same logic for caching Algolia requests.

I guess that depends on what the goal of the timing is: "get the same number as the API"  vs "track how long it takes to return the data".

boringcactus:
Well, the main other source of drift between the two times is the connection pool checkout time, and we've gotten value out of doing spot checks of that in the past, but Finch also [emits Erlang telemetry events](https://hexdocs.pm/finch/Finch.Telemetry.html#module-queue-stop) that we could be using to measure connection pool checkout times directly.

I don't think getting the same number as the V3 API is a goal per se, but having a wildly different number than the V3 API means one of a handful of things is taking too long, and the fewer things it can be, the simpler it is to diagnose. It's not clear that Req supports time measuring for individual steps, but we could stick that into a Nebulex Req plugin if we thought it was worthwhile. (I remember we had discussions about OpenTelemetry and distributed tracing and observability in the past, but it's not clear from crawling through Slack that any of that went anywhere; a [waterfall diagram](https://opentelemetry.io/docs/concepts/observability-primer/#distributed-traces) would be really nice to have for tracking down this sort of performance issue, but it's too late now to have that available for the stuff we're working on now.)

KaylaBrady:
I did recently add [FinchTelemetryHandler](https://github.com/mbta/mobile_app_backend/blob/main/lib/mobile_app_backend/finch_telemetry_handler.ex) for more direct observation of queue vs request times. 

I'd be open to re-writing as a Req plugin but also like that this approach is http client agnostic. Testing Httppoison vs Req was something that helped to hone in on this disk cache issue, so it seems advantageous for it to be easy to switch between alternatives in the near future. @paulswartz what do you think about deferring that for a follow-up task?

paulswartz:
> @paulswartz what do you think about deferring that for a follow-up task?

@KaylaBrady I think that's fine.

KaylaBrady:
👍 ticket created [here](https://app.asana.com/0/1205425564113216/1208390660612728/f)

PR #200, opened 2024-09-20
feat: cache global data instead of recomputing on demand

boringcactus:
### Summary

_Ticket:_ [Cache global data in backend instead of recomputing on demand](https://app.asana.com/0/1205732265579288/1208354000423233/f)

Stores the global data in `:persistent_term` and refreshes it every five minutes. Per the persistent_term docs, we don't need to check if the new data is the same as the old data, because it'll be ignored automatically if it's the same.

### DIFF
--- a/lib/mobile_app_backend/global_data_cache.ex
+++ b/lib/mobile_app_backend/global_data_cache.ex
@@ -33,7 +33,7 @@ defmodule MobileAppBackend.GlobalDataCache do
 
   @spec get_data(key()) :: data()
   def get_data(key \\ default_key()) do
-    :persistent_term.get(key)
+    :persistent_term.get(key, nil) || update_data(key)
   end
 
   @impl GenServer
@@ -44,8 +44,6 @@ defmodule MobileAppBackend.GlobalDataCache do
       key: opts[:key],
       update_ms: opts[:update_ms] || :timer.minutes(5)
     }
-
-    update_data(state.key)
 
     {:ok, state}
   end
@@ -59,7 +57,7 @@ defmodule MobileAppBackend.GlobalDataCache do
     {:noreply, state}
   end
 
-  @spec update_data(key()) :: :ok
+  @spec update_data(key()) :: data()
   defp update_data(key) do
     stops = fetch_stops()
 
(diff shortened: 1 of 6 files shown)


lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -0,0 +1,125 @@
...
+  def default_key, do: __MODULE__
+
+  def start_link(opts) do
+    opts = Keyword.merge([key: default_key()], opts)
+    GenServer.start_link(__MODULE__, opts)
+  end
+
+  @spec get_data(key()) :: data()
+  def get_data(key \\ default_key()) do
+    :persistent_term.get(key)
```
KaylaBrady, line 36: suggestion: Could this first check if the data is populated and call `update_data` if it hasn't been yet? I think making `update_data` more resilient to failure and ensuring the data is present will make this a more reliable data store that seems safer to call outside the global controller flow.

KaylaBrady (commented):
Nice! Great find on not having to compare new & old data. One comment about resiliency, somewhat related to the docker failure

lib/mobile_app_backend/global_data_cache.ex
boringcactus: Ensuring the data is present beforehand was why I'm calling `update_data/1` in `init/1`, so that the GenServer doesn't count as started until the data has been loaded, but I guess if that breaks the Docker container then that approach would work too.

I'm not sure what it would mean for `update_data/1` to be more resilient to failure - returning partial data seems like it'd be incorrect, and if it crashes instead, the `GlobalDataCache` will just restart and retry.

KaylaBrady: Yeah agreed partial data wouldn't make sense. If any of the requests fail I would expect nothing to be written to :persistent_term & maybe an error logged. To your point at planning, we aren't doing that in a lot of places right now, but if-minor, would be nice to modify as part of touching this path.

boringcactus: Well, the error that's logged will be that the calling process crashed, and we'll get a report in Sentry. In practice, the overwhelming majority of the time, `update_data/1` will be called from the `GlobalDataCache` itself, so there's no way for it to respond with an HTTP error.

KaylaBrady (approved):

PR #206, opened 2024-09-25
perf(Predictions.PubSub): Read stops + routes from global cache instead of API

KaylaBrady:
### Summary

What is this PR for?

This PR incorporates the global cache data added in https://github.com/mbta/mobile_app_backend/pull/200 since we found those API calls to be a main source of latency in load testing ([notes](https://www.notion.so/mbta-downtown-crossing/Nearby-Transit-Scalability-Notes-104f5d8d11ea80829652c8f4a0aba748)).


KaylaBrady:
Load testing against this I very quickly started to get errors joining the channel:
```
87955d4a5a6b 13:02:37.701 [error] GenServer #PID<0.8821.0> terminating
87955d4a5a6b ** (MatchError) no match of right hand side value: {:error, %Req.TransportError{reason: :timeout}}
87955d4a5a6b (mobile_app_backend 0.1.0) lib/mobile_app_backend/global_data_cache.ex:194: MobileAppBackend.GlobalDataCache.Impl.fetch_route_patterns/0
87955d4a5a6b (mobile_app_backend 0.1.0) lib/mobile_app_backend/global_data_cache.ex:157: MobileAppBackend.GlobalDataCache.Impl.update_data/1
87955d4a5a6b (mobile_app_backend 0.1.0) lib/mobile_app_backend/predictions/pub_sub.ex:70: MobileAppBackend.Predictions.PubSub.subscribe_for_stops/1
```

I think we will need to either populate the cache on startup or have a fallback mechanism.

KaylaBrady:
Added a call to recalculate as part of init, seeing splunk error though. 
Potentially we should move checking for the presence of the global data into the health check endpoint so that traffic only shifts to an instance once it has global data.

lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -24,40 +26,132 @@ defmodule MobileAppBackend.GlobalDataCache do
     defstruct [:key, :update_ms]
   end
 
-  def default_key, do: __MODULE__
+  @callback default_key :: term()
```
boringcactus, line 29:
```suggestion
  @callback default_key :: key()
```

lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -24,40 +26,132 @@ defmodule MobileAppBackend.GlobalDataCache do
...
     {:ok, state}
   end
 
   @impl GenServer
-  def handle_info(:recalculate, %State{} = state) do
+  def handle_info(:recalculate, %State{update_ms: update_ms} = state) do
     update_data(state.key)
 
-    Process.send_after(self(), :recalculate, state.update_ms)
+    if :persistent_term.get(state.key, nil) do
```
boringcactus, line 144: `update_data` will only not call `:persistent_term.put/2` if it crashes, so I don't think I understand when this check could fail.

lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -24,40 +26,132 @@ defmodule MobileAppBackend.GlobalDataCache do
...
-    opts = Keyword.merge(Application.get_env(:mobile_app_backend, __MODULE__), opts)
+  def init(opts \\ []) do
+    opts = Keyword.merge(Application.get_env(:mobile_app_backend, GlobalDataCache), opts)
 
     state = %State{
       key: opts[:key],
       update_ms: opts[:update_ms] || :timer.minutes(5)
     }
 
+    Process.send_after(self(), :recalculate, :timer.seconds(1))
```
boringcactus, line 135: It's weird that this wasn't here before. This is really tough to test locally, since you can only know it's working if GTFS actually changes, but that might mean this never actually worked and was only ever calculating the global data once. Oops!

lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -24,40 +26,132 @@ defmodule MobileAppBackend.GlobalDataCache do
...
+  def start_link(opts \\ []) do
     opts = Keyword.merge([key: default_key()], opts)
-    GenServer.start_link(__MODULE__, opts)
+    GenServer.start_link(GlobalDataCache, opts)
   end
 
-  @spec get_data(key()) :: data()
+  @impl true
   def get_data(key \\ default_key()) do
     :persistent_term.get(key, nil) || update_data(key)
```
boringcactus, line 97: Maybe this `update_data` should be moved into a `GenServer.call` or something so that if there are a dozen simultaneous calls to `get_data/1` before data is loaded they don't each call `update_data/1`.

boringcactus (commented):
I think being able to use different keys and having a Mox mock are two different ways to solve the problem of providing test data from the GlobalDataCache to consumers, and I'm not quite sure if we need both of them, but I haven't finished thinking through which one would be enough.

If we report that instances are only healthy when they have global data, that'll cause smoke testing the Docker container in CI to fail again, unless we give the Docker container smoke test an API key, which might actually be the correct fix anyway, and I think that would let us actually just fetch the data eagerly.

lib/mobile_app_backend/global_data_cache.ex
KaylaBrady: Added some light tests for this by checking that the message is sent

KaylaBrady:
@boringcactus I like having the Mox available here so that it is possible to mock the higher-level function rather than having to insert all the required data in persistent term. It was especially helpful in the StreamSubscriber tests to be able to mock `route_ids_for_stops` in a way that is consistent with mocking other data.

I don't think giving the docker container an API key would solve the issue - I thought it couldn't make any network requests in CI (these V3 API requests should succeed without an API key anyway)

boringcactus:
That makes sense. If we want to be able to unit test the `GlobalDataCache` itself, we probably still need to be able to run it with arbitrary keys, but for testing things that call it it's more work to make a cache key and fill the persistent data directly. If we're setting up Mox here for that purpose, this is probably the right place to rework the `GlobalControllerTest` to use Mox instead and roll back my changes to `GlobalController` to pick a cache key out of the connection assigns.

I'm not sure I'm aware of any issues with the Docker container not being able to make outgoing network requests in CI - we run load tests in CI against a real API instance, and it works fine. The API requests would succeed with no API key, but they can't succeed with no API URL.

lib/mobile_app_backend/global_data_cache.ex
KaylaBrady: I'd be somewhat worried about putting it into `GenServer.call` since if it is slow for the first user, subsequent user requests will all fail too. 

I think the best bet is making sure that the data is populated first & removing the call to `update_data` from `get_data`. 

Doing that asynchronously via the scheduled checks & preventing user traffic via a healthcheck seems like the safest approach for that to me - if the global data can't load immediately for some reason, it seems cleaner to continue trying to re-fetch without crashing. Maybe I'm overly wary of crashes though. In any case, I think resolving that mechanism could be part of a separate PR so that the polling is in place

KaylaBrady: @boringcactus I'm going to break this PR up into separate ones so that we can clear out the immediate problem of setting up the timed refresh of this data. 

I'm in favor of the health check approach over fetching the data in `init` in failing. In speaking to Paul about it (since he might be pitching in for that change anyway), it has the advantage over the `init` of faster deploys in the case that the request fails. Skate and API take that approach as well.

boringcactus: I don't think Glides has any particular application-logic-specific health checks in its `/_health`, which is probably why I hadn't thought of it, but I guess if that's common we may as well do it here. Good catch on fixing the refresh first - I had completely lost track of that in the context of the TestFlight public beta.

boringcactus (approved):

PR #208, opened 2024-09-27
fix(GlobalDataCache): trigger timed fetch on startup

KaylaBrady:
### Summary

No ticket, a fix found in https://github.com/mbta/mobile_app_backend/pull/206

What is this PR for?
This ensures that the global data cache will periodically check for updated data from the V3 API. Since this involves making a network request shortly after application startup, this also ensures that the docker build has the V3 API URL configured and that it can be skipped during testing.

### DIFF
--- a/config/test.exs
+++ b/config/test.exs
@@ -11,6 +11,8 @@ config :mobile_app_backend, MobileAppBackendWeb.Endpoint,
 # isolated fake processes
 config :mobile_app_backend, start_stream_stores?: false
 
+config :mobile_app_backend, stat_global_cache?: false
+
 # Print only warnings and errors during test
 config :logger, level: :warning
 
--- a/config/test.exs
+++ b/config/test.exs
@@ -11,7 +11,7 @@ config :mobile_app_backend, MobileAppBackendWeb.Endpoint,
 # isolated fake processes
 config :mobile_app_backend, start_stream_stores?: false
 
-config :mobile_app_backend, stat_global_cache?: false
+config :mobile_app_backend, start_global_cache?: false
 
 # Print only warnings and errors during test
 config :logger, level: :warning
--- a/lib/mobile_app_backend/global_data_cache.ex
+++ b/lib/mobile_app_backend/global_data_cache.ex
@@ -1,7 +1,5 @@
 defmodule MobileAppBackend.GlobalDataCache do
   use GenServer
-  alias MBTAV3API.Route
-  alias MBTAV3API.Stop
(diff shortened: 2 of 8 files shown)


KaylaBrady:
cc @boringcactus

config/test.exs
```diff
@@ -11,6 +11,8 @@ config :mobile_app_backend, MobileAppBackendWeb.Endpoint,
 # isolated fake processes
 config :mobile_app_backend, start_stream_stores?: false
 
+config :mobile_app_backend, stat_global_cache?: false
```
boringcactus, line 14: ..should this be `start_global_cache?`?

lib/mobile_app_backend/global_data_cache.ex
```diff
@@ -21,43 +23,105 @@ defmodule MobileAppBackend.GlobalDataCache do
...
     {:ok, state}
   end
 
   @impl GenServer
-  def handle_info(:recalculate, %State{} = state) do
+  def handle_info(:recalculate, %State{update_ms: update_ms} = state) do
     update_data(state.key)
 
-    Process.send_after(self(), :recalculate, state.update_ms)
+    if :persistent_term.get(state.key, nil) do
```
boringcactus, line 112: I still don't think this `if` can ever fail.

boringcactus:
Also, is this the right place to add the global data to `/_health` or should that wait for #206?

lib/mobile_app_backend/global_data_cache.ex
KaylaBrady: Oh right I meant to back that out too, thanks!

KaylaBrady:
I split out a [separate ticket](https://app.asana.com/0/1205732265579288/1208422712282314/f) for the health endpoint and will hold off on the predictions changes until that is done. As-is this hasn't been a problem in prod (yet)

lib/mobile_app_backend/global_data_cache.ex
KaylaBrady: Oh wait this if can fail - it will fail the first time `:recalculate` is called since `update_data` isn't called from `init`

boringcactus: How will that make it fail? `update_data` (which calls `:persistent_term.put/2`) is called right before this `if`.

KaylaBrady: 🤦‍♀️ I should have looked 1 line higher

boringcactus (approved):

PR #214, opened 2024-10-09
feat: Add new multi-route vehicles channel

EmmaSimon:
### Summary

_Ticket:_ [Green Line vehicles don't load in filtered Stop Details](https://app.asana.com/0/1205732265579288/1207813918309501/f)

For the grouped green line stop pages, we need to connect to a vehicles channel that can return multiple routes at once, but the existing one only supports a single route ID. This adds a new channel which accepts any number of route IDs and returns all vehicles matching them.

### DIFF
--- a/lib/mobile_app_backend_web/channels/vehicles_for_route_channel.ex
+++ b/lib/mobile_app_backend_web/channels/vehicles_for_route_channel.ex
@@ -2,6 +2,7 @@ defmodule MobileAppBackendWeb.VehiclesForRouteChannel do
   use MobileAppBackendWeb, :channel
 
   alias MBTAV3API.JsonApi
+  alias MBTAV3API.Stream
   alias MBTAV3API.Vehicle
 
   @throttle_ms 500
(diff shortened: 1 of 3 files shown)


boringcactus (commented):
Why a separate topic rather than accepting a list in `route_id` in the existing topic?

lib/mobile_app_backend_web/channels/vehicles_for_route_channel.ex
```diff
@@ -15,12 +15,34 @@ defmodule MobileAppBackendWeb.VehiclesForRouteChannel do
...
+         direction_id: direction_id,
+         throttler: throttler
+       )}
+    else
+      :error -> {:error, %{code: :missing_param}}
+    end
+  end
+
+  @impl true
+  def join("vehicles:routes:" <> topic_param_concat, _payload, socket) do
```
EmmaSimon, line 33: > Why a separate topic rather than accepting a list in `route_id` in the existing topic?

@boringcactus  @KaylaBrady and I talked in the office yesterday, and we decided that in general we should move towards putting params into the topic names, to make it possible to connect to the same channel with different params if we need to.

boringcactus (approved):

PR #225, opened 2024-10-24
perf(VehiclesForRouteChannel): Use vehicles from PubSub instead of GenServer

KaylaBrady:
### Summary

_Ticket:_ [[Name](https://app.asana.com/0/1205732265579288/1208619681848095/f)](https://app.asana.com/0/1205732265579288/1208619681848095/f)

What is this PR for?

Addresses the bottleneck found in Load Testing when many users are joining vehicles channels at once. 

`Vehicles.PubSub` is largely a copy of `Predictions.PubSub` - there is probably a good basis for extracting out common pieces, but I think it is worth saving that for a separate ticket. 

This does switch `VehicleChannel` to use PubSub as well for consistency.

Testing:
* Added & updated unit tests
* Ran locally on phone pointing to dev-orange deployiment
* Confirmed logs for new PubSub module in Splunk as expected

### DIFF
--- a/lib/mobile_app_backend_web/channels/predictions_for_stops_v2_channel.ex
+++ b/lib/mobile_app_backend_web/channels/predictions_for_stops_v2_channel.ex
@@ -31,14 +31,7 @@ defmodule MobileAppBackendWeb.PredictionsForStopsV2Channel do
   @spec handle_info({:new_predictions, any()}, Phoenix.Socket.t()) ::
           {:noreply, Phoenix.Socket.t()}
   def handle_info({:new_predictions, new_predictions_for_stop}, socket) do
-    {time_micros, _result} =
-      :timer.tc(fn ->
-        :ok = push(socket, "stream_data", new_predictions_for_stop)
-      end)
-
-    Logger.info("#{__MODULE__} push duration=#{time_micros / 1000}")
-
-    require Logger
+    :ok = push(socket, "stream_data", new_predictions_for_stop)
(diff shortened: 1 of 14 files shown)


lib/mobile_app_backend_web/channels/predictions_for_stops_v2_channel.ex
```diff
@@ -31,14 +31,7 @@ defmodule MobileAppBackendWeb.PredictionsForStopsV2Channel do
   @spec handle_info({:new_predictions, any()}, Phoenix.Socket.t()) ::
           {:noreply, Phoenix.Socket.t()}
   def handle_info({:new_predictions, new_predictions_for_stop}, socket) do
-    {time_micros, _result} =
```
KaylaBrady, line 34: tiny log cleanup while copying over behavior of predictions channel

EmmaSimon (approved):
Did you bother running load testing again with these changes just to get a sense of the new upper threshold for connections? It's fine if not, I'm mostly just curious.

KaylaBrady:
@EmmaSimon thanks for asking - definitely worth running again, and I hadn't yet! Added results to the [Load testing doc](https://www.notion.so/mbta-downtown-crossing/Load-Testing-6754b91c079b4c998f6000d9f46af9a2) - this got us to push past 3,300 users at a rate of 5 users / second. I think we'll still want to tackle static data caching to improve that further, or start with a higher base # of instances running.

PR #231, opened 2024-11-01
perf(ScheduleController): Fetch schedules by individual stop for more cache hits

KaylaBrady:
### Summary

_Ticket:_ [Caching schedules data [backend only]](https://app.asana.com/0/1205425564113216/1208639745611053/f)

What is this PR for?

As implemented, if any request for a stop fails, the entire request fails. This is to preserve the existing all-or-nothing behavior for fetching schedules for all stops together. 

During load testing ([notes](https://www.notion.so/mbta-downtown-crossing/Load-Testing-6754b91c079b4c998f6000d9f46af9a2)), there were still big discrepancies between the duration logged for a request from the V3 API vs the duration logged to receive on our backend (602ms vs 4.99s). This PR addresses that problem in the lowest effort way: Adding TTL-based caching to the other static endpoints too to limit the number of concurrent requests to the API. 

With these changes (and load testing script changes from https://github.com/mbta/mobile_app_backend/pull/236), there were a handful of failures when fetching schedules due to timeouts, which still represent a low % of overall requests. There is more optimization we can do here, but this PR feels like a meaningful enough improvement on its own.
image


config/config.exs
```diff
@@ -20,6 +20,10 @@ config :mobile_app_backend, MBTAV3API.ResponseCache,
   gc_interval: :timer.hours(1),
   allocated_memory: 250_000_000
 
+config :mobile_app_backend, MBTAV3API.RepositoryCache,
+  gc_interval: :timer.hours(2),
+  allocated_memory: 250_000_000
```
KaylaBrady, line 25: Setting to 2 hours based on the recommendation [here](https://hexdocs.pm/nebulex/Nebulex.Adapters.Local.html#module-caveats-when-using-ttl-option) that ttl should be less than gc_interval

lib/mbta_v3_api/repository.ex
```diff
@@ -67,24 +67,34 @@ end
...
+
   alias MBTAV3API.JsonApi
 
+  @ttl :timer.hours(1)
+
   @impl true
   def alerts(params, opts \\ []), do: all(MBTAV3API.Alert, params, opts)
 
   @impl true
+  @decorate cacheable(cache: MBTAV3API.RepositoryCache, on_error: :nothing, opts: [ttl: @ttl])
```
KaylaBrady, line 81: pulled from [dotcom](https://github.com/mbta/dotcom/blob/308ca377c3955f64c13e7ac8244bef14131d111c/lib/routes/repo.ex#L6)

lib/mobile_app_backend_web/controllers/schedule_controller.ex
```diff
@@ -39,20 +56,50 @@ defmodule MobileAppBackendWeb.ScheduleController do
...
+          %{schedules: [MBTAV3API.Schedule.t()], trips: JsonApi.Object.trip_map()} | :error
+  defp fetch_schedules_parallel(filters) do
+    filters
+    |> Task.async_stream(fn filter_params ->
+      {filter_params, fetch_schedules(filter_params)}
+    end)
+    |> Enum.reduce_while(%{schedules: [], trips: %{}}, fn result, acc ->
+      case result do
+        {:ok, {_params, %{schedules: schedules, trips: trips}}} ->
+          {:cont, %{schedules: acc.schedules ++ schedules, trips: Map.merge(acc.trips, trips)}}
```
boringcactus, line 81: This concatenation means that the response will no longer properly be sorted by departure time. If that sort order is required downstream, we should preserve it, and if it is not, it might make sense to remove it from `fetch_schedules/1`.

lib/mobile_app_backend_web/controllers/schedule_controller.ex
```diff
@@ -39,20 +56,50 @@ defmodule MobileAppBackendWeb.ScheduleController do
...
+    [stop: stop_id, date: service_date]
+  end
+
+  @spec fetch_schedules_parallel([[JsonApi.Params.filter_param()]]) ::
+          %{schedules: [MBTAV3API.Schedule.t()], trips: JsonApi.Object.trip_map()} | :error
+  defp fetch_schedules_parallel(filters) do
+    filters
+    |> Task.async_stream(fn filter_params ->
+      {filter_params, fetch_schedules(filter_params)}
+    end)
```
boringcactus, line 77: Would it make sense to call this with `ordered: false`?

lib/mbta_v3_api/repository.ex
```diff
@@ -67,24 +67,34 @@ end
...
   def stops(params, opts \\ []), do: all(MBTAV3API.Stop, params, opts)
 
   @impl true
+  @decorate cacheable(cache: MBTAV3API.RepositoryCache, on_error: :nothing, opts: [ttl: @ttl])
   def trips(params, opts \\ []), do: all(MBTAV3API.Trip, params, opts)
 
   @spec all(module(), JsonApi.Params.t(), Keyword.t()) ::
```
boringcactus, line 100: Would it make more sense to only `@decorate cacheable` on `all/3` than on almost all of the individual resources separately?

KaylaBrady: Good point!

lib/mobile_app_backend_web/controllers/schedule_controller.ex
KaylaBrady: The list of schedules is ultimately converted to a map [here](https://github.com/mbta/mobile_app/blob/0832cbc72be3c5a3bf7389ae7c564efe0255127c/shared/src/commonMain/kotlin/com/mbta/tid/mbta_app/model/UpcomingTrip.kt#L104), so it seems safe to remove the sorting. Will run the app locally against dev-orange to confirm.

boringcactus (approved):

lib/mbta_v3_api/repository.ex
EmmaSimon: Do we want to cache alerts though? I can see it potentially being important to _not_ cache them.

KaylaBrady: The alerts fn wasn't actually used since we use the Alerts stream instead - removed in [83e238b](https://github.com/mbta/mobile_app_backend/pull/231/commits/83e238bc328957767c9de51de2f0d9f11c4fef27)

PR #235, opened 2024-11-05
feat: Make trip channel pull from ETS predictions

EmmaSimon:
### Summary

_Ticket:_ [Fix trip details crash when trip not in service](https://app.asana.com/0/1205732265579288/1207687497283289/f)

Update the `predictions:trip:<trip_id>` channel to use the same ETS table that predictions by stop are pulling from. This also better handles the case where the frontend passes an added trip ID that doesn't exist in the trips endpoint, which can happen frequently for vehicles near the end of a line. Now instead of crashing on the backend, an error is returned to the frontend.

Added tests for new behavior


lib/mobile_app_backend/predictions/pub_sub.ex
```diff
@@ -4,20 +4,30 @@ defmodule MobileAppBackend.Predictions.PubSub.Behaviour do
...
   @type predictions_for_stop :: %{Stop.id() => JsonApi.Object.full_map()}
 
-  @type subscribe_response :: %{
+  @type subscribe_stop_response :: %{
           predictions_by_stop: %{Stop.id() => %{Prediction.id() => Prediction.t()}},
           trips: %{Trip.id() => Trip.t()},
           vehicles: %{Vehicle.id() => Vehicle.t()}
         }
 
+  @type subscribe_trip_response :: %{
```
KaylaBrady, line 13: suggestion(non-blocking): add the new `trip_id` field here

lib/mobile_app_backend_web/channels/predictions_for_trip_channel.ex
```diff
@@ -1,74 +1,48 @@
...
-      end
+    Logger.info("#{__MODULE__} push duration=#{time_micros / 1000}")
 
-    %{
-      JsonApi.Object.to_full_map([])
-      | predictions: predictions,
-        trips: Map.take(route_data.trips, [trip_id]),
-        vehicles: Map.take(route_data.vehicles, vehicle_ids)
-    }
+    require Logger
```
KaylaBrady, line 45: nitpick: stray `require Logger`

lib/mobile_app_backend_web/channels/user_socket.ex
```diff
@@ -11,6 +11,7 @@ defmodule MobileAppBackendWeb.UserSocket do
   channel "predictions:stops", MobileAppBackendWeb.PredictionsForStopsChannel
   channel "predictions:stops:v2:*", MobileAppBackendWeb.PredictionsForStopsV2Channel
 
+  channel "predictions:trip:v2:*", MobileAppBackendWeb.PredictionsForTripV2Channel
```
KaylaBrady, line 14: suggestion: this can be removed

KaylaBrady (approved):
Nice work! A few tiny comments

lib/mobile_app_backend/predictions/pub_sub.ex
EmmaSimon: Sorry, I don't understand what you mean?

KaylaBrady: There is a trip_id field added to the map [here](https://github.com/mbta/mobile_app_backend/pull/235/files#diff-03648ca268be035bc968db4cbf1e34012cfd2fba57e08254a1bbc6f47a444957R235), I think it would be worth adding to the typespec

EmmaSimon: Ohh, okay yes, updated

PR #243, opened 2024-11-19
feat(AlertsChannel): Subscribe to PubSub instead of reading from GenServer

KaylaBrady:
### Summary

_Ticket:_ [Remove alert streaming bottleneck](https://app.asana.com/0/1205732265579288/1208690341598921/f)

What is this PR for?

This PR moves from storing the state of alerts in a GenServer to storing it in an ETS so that subscribers can read in parallel. Updates are published to subscribers every 0.5 seconds.

I timeboxed doing some reduction of code duplication here using a macro, but couldn't quite get it to work and the duplication seems minimal enough to not block getting this change out over.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -10,6 +10,7 @@ import Config
 config :mobile_app_backend,
   generators: [timestamp_type: :utc_datetime]
 
+config :mobile_app_backend, alerts_broadcast_interval_ms: 500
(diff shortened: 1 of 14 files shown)


boringcactus (approved):
I'd like to get that refactoring done eventually so important logic doesn't blend in with boilerplate, but that doesn't have to block this PR.

PR #244, opened 2024-11-20
refactor: common PubSub functions + macro

KaylaBrady:
### Summary

Follow up to https://github.com/mbta/mobile_app_backend/pull/243

What is this PR for?

This PR pulls out some common functions from our 3 realtime PubSub modules. I initially tried to do this entirely using macros, but pulling more into regular functions winded up being easier and I think easier to follow.


lib/mobile_app_backend/pub_sub.ex
```diff
@@ -0,0 +1,120 @@
...
+      # consumers don't have to wait `:broadcast_interval_ms` to receive their first message.
+      @impl true
+      def handle_info(:reset_event, state) do
+        send(self(), :broadcast)
+        {:noreply, state, :hibernate}
+      end
+
+      def handle_info(:timed_broadcast, state) do
+        send(self(), :broadcast)
+        interval = unquote(broadcast_interval_ms)
```
boringcactus, line 110: I'm surprised that this `unquote` works, since `broadcast_interval_ms` is not a variable within the scope of `__using__/1` - I'd expect a bare `broadcast_interval_ms` to work here, although storing `broadcast_interval_ms` as a module attribute `@broadcast_interval_ms` might be closer to what we'd write by hand.

lib/mobile_app_backend/predictions/pub_sub.ex
```diff
@@ -45,22 +45,21 @@ defmodule MobileAppBackend.Predictions.PubSub do
 
   Based on https://github.com/mbta/dotcom/blob/main/lib/predictions/pub_sub.ex
   """
-  use GenServer
+  use MobileAppBackend.PubSub,
+    broadcast_interval_ms:
+      Application.compile_env(:mobile_app_backend, :predictions_broadcast_interval_ms, 10_000),
+    broadcast_message_name: :new_vehicles
```
boringcactus, line 51: `broadcast_message_name` doesn't appear to be used within `PubSub.__using__/1`, so it doesn't need to be specified here and in `Vehicles.PubSub` (plus this is the wrong value anyway).

KaylaBrady: Ah yep that is a holdover from trying to stuff more pieces into the macro. Will remove!

KaylaBrady: Ah yep that is a holdover from trying to stuff more pieces into the macro. Will remove!

lib/mobile_app_backend/pub_sub.ex
KaylaBrady: I tried removing unquote but it doesn't work. Isn't `broadcast_interval_ms` defined as a variable in the scope of `__using__/1`/ [here](https://github.com/mbta/mobile_app_backend/pull/244/files#diff-f72fb54bb8f4d6600ef1e7dbdce4949fd1f712ad51295b7fe6097b2a25979198R97)?

I like the approach of passing it as a param so that it helps to direct a reader to see that the interval-based broadcasting logic is done by the macro, rather than having to otherwise search for usage of `@broadcast_interval_ms`

KaylaBrady: Ah thanks for the explanation, I misunderstood your initial suggestion. Will update!

boringcactus (approved):

PR #246, opened 2024-11-21
fix: token rotation flaky tests

KaylaBrady:
### Summary

What is this PR for?

No ticket, small changes to fix 2 flaky tests that seem to be causing more[ CI failures ](https://github.com/mbta/mobile_app_backend/actions/runs/11956483723/job/33331368175)as of late. 

Also a tiny style change to alphabetize the mocks


test/mobile_app_backend/mapbox_token_rotator_test.exs
```diff
@@ -137,6 +137,6 @@ defmodule MobileAppBackend.MapboxTokenRotatorTest do
 
     actual_interval = second_token_time - first_token_time
 
-    assert (actual_interval - refresh_interval) in -5..5
+    assert (actual_interval - refresh_interval) in -50..50
```
KaylaBrady, line 140: Within 50ms of the expected message receive time seems acceptable. Running this a bunch locally, the failures were typically in the range of 8-12ms. I don't think the exact timing matters for this test, it is more important that the message was received twice and the value of the token changed.

EmmaSimon (approved):

PR #247, opened 2024-11-25
fix(RepositoryCache): Reduce cache key collisions

KaylaBrady:
### Summary

No ticket, hotfix for caching issue raised[ in sentry](https://mbtace.sentry.io/issues/6089125911/events/de7517128def48b2bba18b8c9b768583/?project=4506781622403072).

What is this PR for?

Sentry recorded an error where hitting `/api/stop/map?stop_id=place-north` was crashing while trying to process schedule records for the 69 bus. The 69 bus doesn't go to north station, which made this extra suspicious. 

 I was able to reproduce this locally by 
1. Hitting /api/schedules?stop_ids=1419&date_time=2024-11-25T10:45:52-05:00
2. Hitting /api/stop/map?stop_id=place-north

The cause appears to be a cache key collision. Right now Nebulex is using the [SimpleKeyGenerator](https://hexdocs.pm/nebulex/Nebulex.Caching.SimpleKeyGenerator.html), which does only a hash of the given arguments.

For comparison, dotcom uses a [custom key generator](https://github.com/mbta/dotcom/blob/main/lib/dotcom/cache/key_generator.ex) which includes the function name in addition to the hashed arguments. 

This switches to use the same custom cache key function as dotcom (less some module name reformatting pieces). By moving the cache decorator to each function in Repository, there can never be an instance of cache key collisions across data types. 

There could potentially still be key collisions within a single data type - we could include the entire unhashed function args, or switch to a different hashing approach, but I expect that name-spacing the cache key by function will help substantially.


lib/mbta_v3_api/repository.ex
```diff
@@ -60,28 +60,32 @@ defmodule MBTAV3API.Repository.Impl do
...
   def stops(params, opts \\ []), do: all(MBTAV3API.Stop, params, opts)
 
   @impl true
+  @decorate cacheable(cache: RepositoryCache, on_error: :nothing, opts: [ttl: @ttl])
   def trips(params, opts \\ []), do: all(MBTAV3API.Trip, params, opts)
 
   @spec all(module(), JsonApi.Params.t(), Keyword.t()) ::
           {:ok, JsonApi.Response.t(JsonApi.Object.t())} | {:error, term()}
-  @decorate cacheable(cache: MBTAV3API.RepositoryCache, on_error: :nothing, opts: [ttl: @ttl])
   defp all(module, params, opts) do
```
boringcactus, line 89: Since the first argument here is different for every type of query being made, I would expect that keeping the decorator on `all/3` would be fine - perhaps since there are no default arguments at play that would let us do something like
```elixir
@impl Nebulex.Caching.KeyGenerator
def generate(_, _, [module, params, opts]) do
  "#{module}|#{:erlang.phash2({params, opts})}"
end
```
Maybe that's too clever, though, and it still won't work and we'll be confused trying to debug it.

lib/mbta_v3_api/repository_cache/key_generator.ex
```diff
@@ -0,0 +1,22 @@
+defmodule MBTAV3API.RepositoryCache.KeyGenerator do
```
boringcactus, line 1: Since this module won't be referenced directly outside of `RepositoryCache`, it might make sense to define it as a nested module directly within `RepositoryCache` instead of giving it its own file and subdirectory.

lib/mbta_v3_api/repository.ex
KaylaBrady: Yeah that does seem like it would be equivalent. I do think it is beneficial to have the decorator on each function name so it is clearer when looking just at the repository module that caching is done at the function-level without having to dig into the key generation

lib/mbta_v3_api/repository_cache/key_generator.ex
boringcactus: Oh, what I meant was
```elixir
defmodule MBTAV3API.RepositoryCache do
  defmodule KeyGenerator do
  end
end
```
but this seems reasonable too.

boringcactus (approved):

PR #264, opened 2025-01-08
fix: reject nil in Vehicle.t direction_id

boringcactus:
### Summary

_Ticket:_ none

This doesn't actually cause any issues in the backend, but it causes issues in the frontend, and now that #272 is in place we can easily discard vehicles with no direction.


KaylaBrady (dismissed):
Can you make a ticket to revisit these changes once [the upstream fix](https://app.asana.com/0/1204137169527258/1209119722014888/f) is done?

KaylaBrady:
@boringcactus I noticed in the linked ticket that null direction_id is a known case for vehicles on a shuttle. This filtering might be too aggressive if there is in fact a valid use case for null direction_id. With that, I think the previous approach of documenting that null is an acceptable value would be appropriate.

boringcactus:
Per Paul S, shuttles have a null direction_id because they aren't assigned to trips, but that also means they can't have predictions, and if we started to have shuttle predictions then shuttles would have trips and therefore direction_ids. Since the only ways to load a vehicle are via filtering by direction and via opening a prediction, and neither of those can load shuttles, we will never show shuttles anyway and so we can just discard them explicitly instead of handling the null case in the frontend.

PR #271, opened 2025-02-04
fix: give most enums default values

boringcactus:
### Summary

_Ticket:_ [Investigate: Missing alerts in prod iOS app](https://app.asana.com/0/1205732265579288/1209298415313011)

I remember wanting this in #204 but not being able to figure out a good approach. I'm not sure why I didn't think of this.

### DIFF
--- a/lib/mbta_v3_api/schedule.ex
+++ b/lib/mbta_v3_api/schedule.ex
@@ -17,7 +17,8 @@ defmodule MBTAV3API.Schedule do
 
   Util.declare_enum(
     :stop_edge_type,
-    Util.enum_values(:index, [:regular, :unavailable, :call_agency, :coordinate_with_driver])
+    Util.enum_values(:index, [:regular, :unavailable, :call_agency, :coordinate_with_driver]),
+    :call_agency
   )
 
   @derive Jason.Encoder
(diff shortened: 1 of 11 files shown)


lib/mbta_v3_api/schedule.ex
```diff
@@ -17,7 +17,8 @@ defmodule MBTAV3API.Schedule do
 
   Util.declare_enum(
     :stop_edge_type,
-    Util.enum_values(:index, [:regular, :unavailable, :call_agency, :coordinate_with_driver])
+    Util.enum_values(:index, [:regular, :unavailable, :call_agency, :coordinate_with_driver]),
+    :call_agency
```
KaylaBrady, line 21: question: This is described in the swagger docs as "Must phone agency to arrange pickup" which doesn't seem like a great default to me (though I don't think we use this anywhere client-side right now). Since this field isn't nullable and has well defined types, why not fall back to `CrashOnUnknownEvenThoughThisIsABadIdea` for any required / well defined GTFS field like is done already for `Route`? I could also see that being applied to `Stop.location_type` and `Vehicle.current_status`

boringcactus:
As discussed at dev check-in this afternoon, we want a more holistic solution to parsing robustness, which for reasons that made sense at the time I put into a separate PR entirely: #272.

PR #272, opened 2025-02-04
fix: improve parsing robustness

boringcactus:
### Summary

_Ticket:_ [Investigate: Missing alerts in prod iOS app](https://app.asana.com/0/1205732265579288/1209298415313011)

Sometimes I miss working in languages with real error handling.

I may be doing too much at once here.

Logical changes:
1. Give some enums inherent default values (subsumes #271 but adds recording observed anomalies in Sentry)
2. Use `!` in parse and serialize functions to indicate crashing on error
3. Discard and log in Sentry unparseable values in static and stream responses
4. Make route types non-nullable


EmmaSimon (approved):
🙌

PR #285, opened 2025-03-03
feat: Alert health checks

EmmaSimon:
### Summary

_Ticket:_ [Automated alerts when stale realtime data](https://app.asana.com/0/1205732265579288/1209334479098774)

In addition to the existing health check for global data cache being populated, also check that the number of alerts in the store is the same as the number coming from the backend.

This includes some significant refactoring for how health checks are handled in general, now we have a `Checker` module, which can be implemented to perform any other checks that we want in the future.

I created a [Splunk alert](https://mbta.splunkcloud.com/en-US/app/search/alert?s=%2FservicesNS%2Fnobody%2Fsearch%2Fsaved%2Fsearches%2FMobile%2520Backend%2520-%2520Healthcheck%2520Failures) that sends a message to #tid-mobile-app-alerts if healthcheck failures are logged more than 15 times in 15 minutes. It doesn't page yet, but once we have a sense of how reliable it is, we'll probably want it to.

### Testing:
 
Added unit tests for new healthcheck behavior, and for the new alert repository.


lib/mobile_app_backend/health/checker/alerts.ex
```diff
@@ -0,0 +1,34 @@
...
+  def healthy? do
+    all_stored = Store.Alerts.fetch([])
+    from_backend = Repository.alerts([])
+
+    case from_backend do
+      {:ok, %{data: from_repo}} ->
+        Checker.log_failure(
+          length(all_stored) == length(from_repo),
+          "stored alert count #{length(all_stored)} != backend alert count #{length(from_repo)}"
+        )
```
boringcactus, line 28: It feels weird to be calling `log_failure` unconditionally, and also, this will perform the string concatenation regardless of whether or not the check fails. I think
```elixir
if length(all_stored) != length(from_repo) do
  Checker.log_failure("stored alert count #{length(all_stored)} != backend alert count #{length(from_repo)}")
end
```
would be a more intuitive design.

EmmaSimon: Would it make more sense if it had a different name, maybe `log_if_unhealthy`/`log_when_unhealthy` or something? I was trying to avoid needing to do the same assigns and ifs in every checker, if we switched to using if statements everywhere, it think it would actually need to look like 
```elixir
case from_backend do
  {:ok, %{data: from_repo}} ->
    healthy = length(all_stored) == length(from_repo)
    if !healthy do
      Checker.log_failure("stored alert count #{length(all_stored)} != backend alert count #{length(from_repo)}")
    end
    healthy

  _ ->
    Checker.log_failure("backend alert request failed")
    false
end
```
which struck me as unnecessarily verbose, especially when repeated across every checker we add.

boringcactus: Ah, yeah, if it needs to both log the message and return the status then that's not actually that easy. I think an architecture like
```elixir
def check_health do
  # ...
  case from_backend do
    {:ok, %{data: from_repo}} ->
      if length(all_stored) == length(from_repo) do
        :ok
      else
        {:error, "stored alert count #{length(all_stored)} != backend alert count #{length(from_repo)}"}
      end

    _ ->
      {:error, "backend alert request failed"}
  end
end
```
might be more idiomatic Elixir, although it'd probably imply replacing `Overall` with something that'll actually sort through and log each failure individually before returning a status to the health endpoint.

boringcactus (approved):

lib/mobile_app_backend/health/checker/alerts.ex
boringcactus: I think this looks pretty good.

PR #286, opened 2025-03-03
chore: change endpoint adapter to bandit

KaylaBrady:
### Summary

_Ticket:_ [Sentry - Ranch listener error...](https://app.asana.com/0/1205732265579288/1209491378620156)

What is this PR for?

We were getting repeated Splunk / Sentry errors for invalid spam calls to our backend - bad endpoints, bad param encoding, bad mime types, etc. 

It was especially noisy & unexpected to see these sentry errors, which should be blocked by [the Sentry.DefaultEventFilter](https://hexdocs.pm/sentry/Sentry.DefaultEventFilter.html). 

@boringcactus had the idea to try switching to use the Bandit adapter to see if that would resolve the issue. And it does! 

Bandit is[ now what is used by default ](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-adapter-configuration) for new phoenix applications and is already used by a number of tid applications, including[ the api ](https://github.com/mbta/api/pull/678)for about 1.5 years.
### Testing
* Confirmed in Splunk / Sentry that while previously making a request with header `'Accept: application/dns-json'` would result in error `Ranch listener MobileAppBackendWeb.Endpoint.HTTP, connection process #PID<0.249240.0>, stream 1 had its request process #PID<0.249241.0> exit with reason {{{%Phoenix.NotAcceptableError{...`, `NotAcceptableError` did not hit the logs or sentry after switching to Bandit.
* Ran app locally against dev-orange. Static requests + phoenix channels worked as expected, no unexpected logs in splunk


KaylaBrady:
@paulswartz - anything else that we should look out for in switching to Bandit from your experience?

boringcactus (commented):
Would it make sense to also remove our `plug_cowboy` dependency?

config/config.exs
```diff
@@ -25,8 +25,7 @@ config :mobile_app_backend, MBTAV3API.RepositoryCache,
 # Configures the endpoint
 config :mobile_app_backend, MobileAppBackendWeb.Endpoint,
   url: [host: "localhost"],
-  http: [compress: true],
```
KaylaBrady, line 28: note: compress is[ true by default ](https://hexdocs.pm/bandit/Bandit.html#t:http_options/0)

paulswartz:
> @paulswartz - anything else that we should look out for in switching to Bandit from your experience?

@KaylaBrady the one thing I noticed was that in some configurations, WebSockets weren't being passed all the way through the load balancer. If you're still seeing them work normally (or you aren't using WebSockets) then it's been straightforward.

KaylaBrady:
Thanks @paulswartz! We are using WebSockets via channels for streaming predictions, alerts, and vehicles, but testing locally against dev-orange those are all still loading in the app as expected.

PR #302, opened 2025-03-28
build: schedule weekly deploys to dev-orange if it’s not in use

boringcactus:
### Summary

_Ticket:_ [iOS & Android dev orange apps](https://app.asana.com/0/1205732265579288/1208842736619571/f)

This seems like it'll probably work.

### DIFF
--- a/.github/workflows/deploy-dev-orange.yml
+++ b/.github/workflows/deploy-dev-orange.yml
@@ -39,7 +39,7 @@ jobs:
   deploy:
     name: Deploy ${{ github.event_name == 'pull_request' && 'PR' || github.ref_type }} ${{ github.event_name == 'pull_request' && github.event.pull_request.number || github.ref_name }} to Dev-Orange
     needs: check
-    if: needs.check.outputs.should-run
+    if: ${{ needs.check.outputs.should-run == 'true' }}
     uses: ./.github/workflows/deploy.yml
     with:
       env: dev-orange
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-dev-orange.yml
```diff
@@ -1,17 +1,47 @@
...
 on:
   workflow_dispatch:
+    inputs:
+      ack:
+        description: "Manual deploys to dev-orange will be wiped out weekly; consider using a PR instead"
+        required: true
+        type: choice
+        options:
+          - I am just deploying main anyway
+          - I am for some other reason okay with that
```
KaylaBrady: Nice confirmation

boringcactus, line 13: I'm hoping that `required: true` with no `default:` will force an explicit selection of one or the other of these before trying to run the workflow. If it doesn’t, I think it'd be more confusing to fail the workflow if this wasn’t acknowledged, so I’ll probably leave a message here in `inputs` with no actual meaning.

boringcactus: It does not; with no `default:` a `choice` will default to the first option.

KaylaBrady (approved):

boringcactus:
Despite printing `should-run=false`, the deployment appears to have run anyway.

boringcactus:
It seems like `'false'` might be truthy. I do not like GitHub Actions.

PR #311, opened 2025-05-01
feat(Health.Checker.Alerts): Evaluate within grace period

KaylaBrady:
### Summary

What is this PR for?
No ticket, this is in response to some [recent noisy health check count pages](https://mbta.slack.com/archives/C01B0577NH4/p1746030439438929) and is intended to avoid noisy failing health check pages in the future.

Previously, we were checking if alert data is stale by comparing the count of alerts in our store to the count of alerts returned from the API. We knew this was a first iteration that would be flaky since it can take a bit of time for our Store to catch up, which causes frequent health check failures (65 in the last 4 hours on prod). So far this hasn't resulted in any pages, but it does create some noise in our dashboard and has the potential to erroneously page someone external to our team, who wouldn't be able to immediately see this as noise. 

This PR adds a grace period of 5 minutes to our alert count check to reduce health check failure noisiness. 

## Testing
* Added unit tests
* Deployed to dev-orange to hopefully evaluate this cutting down on failures, but turns out we don't see all that many failures on dev-orange (presumably because alerts change less frequently). Will evaluate on staging for a while once merged.

### DIFF
--- a/test/mobile_app_backend/health/checker/alerts/last_fresh_store_test.exs
+++ b/test/mobile_app_backend/health/checker/alerts/last_fresh_store_test.exs
@@ -1,21 +1,17 @@
 defmodule MobileAppBackend.Health.Checker.Alerts.LastFreshStoreTest do
-use ExUnit.Case
-alias MobileAppBackend.Health.Checker.Alerts
+  use ExUnit.Case
+  alias MobileAppBackend.Health.Checker.Alerts
 
-setup do
-        start_link_supervised!(Alerts.LastFreshStore)
-        :ok
-end
+  setup do
+    start_link_supervised!(Alerts.LastFreshStore)
+    :ok
+  end
 
   test "read and write last fresh timestamp" do
-
     Alerts.LastFreshStore.update_last_fresh_timestamp(~U[2024-01-01 00:00:00Z])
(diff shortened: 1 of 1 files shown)


test/mobile_app_backend/health/checker/alerts/last_fresh_store_test
boringcactus, line 1: I think this file might be missing a `.exs`.

lib/mobile_app_backend/health/checker/alerts.ex
```diff
@@ -11,25 +11,50 @@ defmodule MobileAppBackend.Health.Checker.Alerts.Impl do
...
+  require Logger
 
   @behaviour MobileAppBackend.Health.Checker
 
+  @grace_period 300
+
   @impl true
   def check_health do
-    store_count = length(Store.Alerts.fetch([]))
+    store_count = length(Store.Alerts.fetch(fields: [alert: []]))
```
KaylaBrady, line 24: We sometimes have seen timeouts on health checks ([ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1209680445093527?focus=true)). My hypothesis is that since we don't actually need the alert fields here, requesting the sparse data will be faster.

boringcactus (approved):

PR #314, opened 2025-05-09
feat: Add route stops endpoint

EmmaSimon:
### Summary

_Ticket:_ [Favorites | Route details - ordered list of stops endpoint](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816289542478?focus=true)

Add a new endpoint `/api/route/stops` which returns a list of stop ids for each of the stops served along that route. This list will include stops that are only on uncommon route patterns or on different branches, ordered in a sort of coherent way (which is determined by the v3 API).

### DIFF
--- a/test/mobile_app_backend_web/controllers/route_controller_test.exs
+++ b/test/mobile_app_backend_web/controllers/route_controller_test.exs
@@ -1,0 +1,58 @@
+defmodule MobileAppBackendWeb.RouteControllerTest do
+  use HttpStub.Case
+  use MobileAppBackendWeb.ConnCase
+  import Mox
+  import Test.Support.Helpers
+  import MobileAppBackend.Factory
+
+  describe "GET /api/route/stops unit tests" do
+    setup do
+      verify_on_exit!()
+      reassign_env(:mobile_app_backend, MBTAV3API.Repository, RepositoryMock)
+    end
+
+    defp mock_route_data do
+      andrew = build(:stop, id: "andrew")
+      jfk = build(:stop, id: "jfk/umass")
+      savin = build(:stop, id: "savin_hill")
+      north_quincy = build(:stop, id: "north_quincy")
+
+      RepositoryMock
+      |> expect(:stops, 1, fn params, _opts ->
+        case params
+             |> Keyword.get(:filter) do
+          [route: "Red", direction_id: "0"] ->
+            ok_response([andrew, jfk, savin, north_quincy])
+        end
+      end)
+
+      RepositoryMock
+      |> expect(:stops, 1, fn params, _opts ->
+        case params
+             |> Keyword.get(:filter) do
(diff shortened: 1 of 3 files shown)


test/mobile_app_backend_web/controllers/route_controller_test.exs
```diff
@@ -0,0 +1,58 @@
...
+      end)
+
+      RepositoryMock
+      |> expect(:stops, 1, fn params, _opts ->
+        case params
+             |> Keyword.get(:filter) do
+          [route: "Red", direction_id: "1"] ->
+            ok_response([north_quincy, savin, jfk, andrew])
+        end
+      end)
```
boringcactus, line 36: Point 4 in the [Mox docs](https://hexdocs.pm/mox/Mox.html) suggests that it’s more idiomatic to do something like
```elixir
RepositoryMock
|> expect(:stops, 2, fn
  [filter: [route: "Red", direction_id: "0"], fields: _], _opts -> ok_response([andrew, jfk, savin, north_quincy])
  [filter: [route: "Red", direction_id: "1"], fields: _], _opts -> ok_response([north_quincy, savin, jfk, andrew])
end)
```
rather than stacking multiple `expect` calls, although if this pattern works and we’re already doing it elsewhere then it’s probably fine.

boringcactus (commented):
For grouped lines, it looks like the V3 API will let you pass `?filter[route]=Green-B,Green-C,Green-D,Green-E`; do we want to accept `?line_id=line-Green` in this endpoint and look up the routes in the backend, or do we want to request `?route_id=Green-B,Green-C,Green-D,Green-E` in the frontend and just pass that through transparently?

EmmaSimon:
> For grouped lines, it looks like the V3 API will let you pass `?filter[route]=Green-B,Green-C,Green-D,Green-E`; do we want to accept `?line_id=line-Green` in this endpoint and look up the routes in the backend, or do we want to request `?route_id=Green-B,Green-C,Green-D,Green-E` in the frontend and just pass that through transparently?

I think it makes more sense to defer that to the frontend, that gives us more flexibility with how we ultimately implement the GL (like is it split into separate lines with a selector for each one or do we show all the stops all at once), and doing the line resolution on the backend means that we'll need to filter out all of the shuttle routes that are also returned with the line data, which doesn't seem worth it when we'll need to do some level of GL special casing on the frontend anyway and might end up wanting the separate branch level ID lists anyway.

boringcactus:
In that case, it might at least be worth testing that comma-separated lists passed in as `route_id` get forwarded as comma-separated lists to the V3 API.

EmmaSimon:
> In that case, it might at least be worth testing that comma-separated lists passed in as `route_id` get forwarded as comma-separated lists to the V3 API.

It does! But yeah I'll add a test for it.

boringcactus (approved):

PR #318, opened 2025-05-21
feat: use the entire backend for deep links

boringcactus:
### Summary

_Ticket:_ [Create link that opens the app if you have it installed and the store otherwise](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209949928014406?focus=true)

Since having the deep links live within dotcom didn’t work, the deep links may as well live here, and since there’s nothing here that we need people to be able to open outside the app, we can just open any path within the backend as a deep link.

Devops PR to create the environment variables: https://github.com/mbta/devops/pull/2904


EmmaSimon (approved):
Nice work! 🎉

PR #319, opened 2025-05-23
feat: Cleaner timeout handling

KaylaBrady:
### Summary

_Ticket:_ [Sentry - Backend Health Check Endpoint Timeouts](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209680445093527?focus=true)

What is this PR for?
This PR cleans up our two most common backend errors on sentry - timeouts of async tasks in the health endpoint & schedules endpoint.


lib/mobile_app_backend/health/check.ex
```diff
@@ -5,22 +5,31 @@ defmodule MobileAppBackend.Health.Check do
...
   alias MobileAppBackend.Health.Checker
 
+  require Logger
+
   @doc """
   Returns true if all Checkers are healthy.
   """
   @spec healthy?() :: boolean()
-  def healthy? do
+  def healthy?(timeout \\ 10_000) do
```
KaylaBrady, line 14: This increases the timeout from 5s to 10s. Looking at splunk over the last 40 days, there are places where`get_json_response` records 4s+ durations for fetching alerts from the V3 API. Slow V3 API queries to verify if our alerts are out of date shouldn't cause us to report an instance as unhealthy, so bumping the timeout seems worthwhile. 

I also debated removing `Task.await_many` entirely here in favor of just letting AWS determine what timeout is appropriate for the health check but figured it is worth keeping it & logging slowness so that we can more easily detect & optimize problems later if needed.

boringcactus (approved):

PR #328, opened 2025-06-04
feat: collapse adjoining alert active periods

boringcactus:
### Summary

_Ticket:_ [Join consecutive alert periods](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1209816091144021?focus=true)

I dug up the original alert from that Slack thread in prediction_loc, and it’s a good thing I did, because I had assumed this was 3:00:00–3:00:00 or 3:00:00–2:59:59 rather than the 3:00:00–2:59:00 that it is. As such, I’m dumping that entire alert into this test case, just to be certain.


KaylaBrady (approved):
👍 good call pulling the old alert!

PR #331, opened 2025-06-12
feat: Add route filter search endpoint

EmmaSimon:
### Summary

_Ticket:_ [🤖 | Favorites | Add stops flow - search / filter](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210229155855234?focus=true)

This returns a max of 50 results to use to filter the list of displayed routes on the route picker page. The top level search only returns a max of 5 route results.


boringcactus (approved):
I had been envisioning this filtering running within the frontend, but I guess it’d be weird if it behaved differently from the top-level search in the app.

PR #344, opened 2025-07-30
fix: return all route search results

boringcactus:
### Summary

_Ticket:_ [Favorites | QA | route search](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210826290372839?focus=true)

Filtering by route type in Algolia is not currently possible since route types aren’t set up as a facet, but even if they were, that wouldn’t let us distinguish SL buses from non-SL buses; we’d need to add a new `isSilverLine` property on all the routes as they go into the index, and then we’d need to set that property as a facet as well. Filtering by route type in the `SearchController` would also not be able to distinguish SL routes. The frontend already knows which routes are on the SL page, so I think it makes the most sense to just return every single search result from the backend and let the frontend continue to filter based on which routes are included in the page being viewed.

### DIFF
--- a/lib/mobile_app_backend/search/algolia/query_payload.ex
+++ b/lib/mobile_app_backend/search/algolia/query_payload.ex
@@ -31,7 +31,7 @@ defmodule MobileAppBackend.Search.Algolia.QueryPayload do
   def for_route_filter(query) do
     Algolia.Index.index_name(:route)
     |> new(query)
-    |> with_hit_size(50)
+    |> with_hit_size(1000)
   end
 
   defp new(index_name, query) do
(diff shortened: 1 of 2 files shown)


EmmaSimon (approved):
I'm slightly worried about response sizes here, especially since we'll likely often do a query with one or two characters as the user starts typing. It shouldn't be a big deal, and we can address it if/when it becomes a real problem. That said, do you have a sense of how difficult it would be to set up the facets the way that we want?

boringcactus:
I think defining `route.type` as a facet that can be used for filtering should be fairly straightforward, and probably wouldn’t cause any issues for dotcom; injecting an `isSilverLine` would be [not particularly difficult, actually](https://github.com/mbta/dotcom/blob/18160ef928b3fcf20bee35771a4062b0b0a93807/lib/algolia/object.ex#L39-L43), and we could probably instead define the line ID as also a facet and then filter by `route.line_id:line-SLWaterfront OR route.line_id:line-SLWashington`. Mostly I’m just hesitant to mess with stuff that dotcom owns and we’re merely borrowing.

PR #354, opened 2025-08-07
feat: Add Algolia facet filtering

EmmaSimon:
### Summary

_Ticket:_ [Favorites | Use algolia route facets](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210967667899893?focus=true)

Add `type` and `line_id` query params to the `/search/routes` endpoint, which are passed to algolia as facet filters.
Both params can be an arbitrarily long list of comma separated values, and types must be sent in the string format passed between the app and the backend (`bus`, `light_rail`, etc).

A request of `/search/routes?query=term&line_id=line-SLWaterfront,line-SLWashington&type=bus` will send algolia a `filters` param set to `(route.line_id:line-SLWaterfront OR route.line_id:,line-SLWashington) AND (route.type:3)`.

Backwards compatibility is maintained, the new params aren't required or anything, so current staging versions will continue working with these changes.


lib/mobile_app_backend/search/algolia/query_payload.ex
```diff
@@ -7,13 +7,17 @@ defmodule MobileAppBackend.Search.Algolia.QueryPayload do
 
   @type t :: %__MODULE__{
           index_name: String.t(),
-          params: map()
+          params: map(),
+          filters: String.t()
         }
 
+  @type facets :: %{String.t() => String.t()}
```
boringcactus, line 14: What if instead we passed the facets as a `%{String.t() => [String.t()]}`? That’d let us avoid joining with `,` when parsing multiple route types just to split on `,` again.

EmmaSimon: Thanks for the suggestion, I like that better!

PR #359, opened 2025-08-12
 feat(RouteBranching.segment): Switch route branching order

KaylaBrady:
### Summary

_Ticket:_ [ | Favorites | Address UI QA - route branching order](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1211025684993056?focus=true)

What is this PR for?
This PR switches the order of left/right lane assignment when there are multiple parallel segments to assign. 

Updated look at the 15:
imageimageimage

Before/After of the RL
imageimage


lib/mobile_app_backend/route_branching/segment.ex
```diff
@@ -575,7 +575,7 @@ defmodule MobileAppBackend.RouteBranching.Segment do
   defp assign_segments([s], [1, 2], _), do: %{s => 2}
   defp assign_segments([s], [0, 2], _), do: %{s => 0}
   defp assign_segments([s1, s2], [t1, t2], _), do: %{s1 => t1, s2 => t2}
-  defp assign_segments([s1, s2], [0, 1, 2], _), do: %{s1 => 0, s2 => 2}
+  defp assign_segments([s1, s2], [0, 1, 2], _), do: %{s1 => 2, s2 => 0}  #%{s1 => 0, s2 => 2}
```
KaylaBrady, line 578: maybe need to flip all of them?

lib/mobile_app_backend/route_branching/segment.ex
```diff
@@ -574,9 +574,9 @@ defmodule MobileAppBackend.RouteBranching.Segment do
   defp assign_segments([s], [0, 1], _), do: %{s => 0}
   defp assign_segments([s], [1, 2], _), do: %{s => 2}
   defp assign_segments([s], [0, 2], _), do: %{s => 0}
-  defp assign_segments([s1, s2], [t1, t2], _), do: %{s1 => t1, s2 => t2}
-  defp assign_segments([s1, s2], [0, 1, 2], _), do: %{s1 => 0, s2 => 2}
-  defp assign_segments([s1, s2, s3], [0, 1, 2], _), do: %{s1 => 0, s2 => 1, s3 => 2}
+  defp assign_segments([s1, s2], [t1, t2], _), do: %{s1 => t2, s2 => t1}
+  defp assign_segments([s1, s2], [0, 1, 2], _), do: %{s1 => 2, s2 => 0}
+  defp assign_segments([s1, s2, s3], [0, 1, 2], _), do: %{s1 => 2, s2 => 1, s3 => 0}
```
KaylaBrady, line 579: should the single segment [0,2] case be flipped as well?
Did the 3 segment case need to be flipped? I think so for since hingham/hull looks good this way

KaylaBrady: Maybe should just sort the `unconstrained_segments` being passed to `assign_segments`

boringcactus: May as well flip the `[s], [0, 2]`, yeah. The 3 segment case is rare, and if Hingham/Hull has it and looks better flipped, then flipping it is correct.

PR #363, opened 2025-08-15
feat: Add T-Alert URL campaign tracking params

EmmaSimon:
### Summary

_Ticket:_ [Add campaign tracking to T-alerts CTA on bus delay alerts](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1210919282215092?focus=true)

Add a new `go.mbta.com/t-alert` endpoint, which will deeplink into the app if on a device with the app already installed, or redirect to the dotcom `app-store` endpoint if not, including the T Alerts link campaign params so that we can tell when users are installing the app from that link.

After this is deployed, we'll need a follow up alerts-concierge change to update the URL. This also does not include GA events within the app if opened with a deep link (see [ticket](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1211066386453647?focus=true)).

### DIFF
--- a/lib/mobile_app_backend_web/controllers/deep_link_controller.ex
+++ b/lib/mobile_app_backend_web/controllers/deep_link_controller.ex
@@ -1,7 +1,7 @@
 defmodule MobileAppBackendWeb.DeepLinkController do
   use MobileAppBackendWeb, :controller
 
-  defp t_alert_cta_campaign_params(),
+  defp t_alert_cta_campaign_params,
     do: %{
       "pt" => "117998862",
       "ct" => "TAlerts",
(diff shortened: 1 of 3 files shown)


lib/mobile_app_backend_web/controllers/deep_link_controller.ex
```diff
@@ -1,12 +1,29 @@
 defmodule MobileAppBackendWeb.DeepLinkController do
   use MobileAppBackendWeb, :controller
 
+  defp t_alert_cta_campaign_params(),
+    do: %{
```
boringcactus, line 5: I only prefer `def foo, do: bar` over `def foo do bar end` when the former actually fits on one line; in cases like this I think a `do ... end` block is cleaner than the explicit keyword list `, do:`.

boringcactus (approved):

PR #372, opened 2025-08-25
fix: Update TAlerts CTA campaign params

EmmaSimon:
### Summary

_Ticket:_ No ticket

The TAlerts CTA link is not properly tracking the campaign when it redirects to dotcom, though based on Splunk logs, we are getting traffic to it with the params set here. This is an attempt to fix it at least for Android, other GA campaign URLs on dotcom seem to follow this format. It's also not working for iOS though, which is surprising because those params seem definitely correct.


boringcactus (approved):
🤞

PR #453, opened 2026-04-07
feat: log User-Agent header and request ID in more places

boringcactus:
### Summary

_Ticket:_ [Add platform and app version to Splunk logs](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1213889644889727)

Pairs with https://github.com/mbta/mobile_app/pull/1647.

I looked through the controllers and the only inter-process handoff is the parallel schedules fetch; I think all our stores are read from channels, and those don’t really have request IDs in the same way.

### DIFF
--- a/lib/mbta_v3_api/headers.ex
+++ b/lib/mbta_v3_api/headers.ex
@@ -8,7 +8,21 @@ defmodule MBTAV3API.Headers do
   @spec build(String.t() | nil) :: header_list
   def build(api_key) do
     []
+    |> derived_request_id()
     |> api_key_header(api_key)
+  end
+
+  @spec derived_request_id(header_list()) :: header_list()
+  defp derived_request_id(headers) do
+    Logger.metadata()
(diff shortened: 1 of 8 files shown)


lib/mbta_v3_api/headers.ex
```diff
@@ -8,9 +8,23 @@ defmodule MBTAV3API.Headers do
...
     |> api_key_header(api_key)
   end
 
+  @spec derived_request_id(header_list()) :: header_list()
+  defp derived_request_id(headers) do
+    Logger.metadata()
+    |> Keyword.fetch(:request_id)
+    |> case do
+      {:ok, request_id} ->
+        [{"X-Request-Id", "#{request_id}/#{System.unique_integer()}"} | headers]
```
boringcactus, line 21: I checked that this request ID is used in the V3 API logs; this should make detailed distributed tracing more straightforward.

KaylaBrady (approved):
🙌

PR #457, opened 2026-04-10
fix: Future alert reminder check

EmmaSimon:
### Summary

_Ticket:_ [Notifications QA | no notifications sent for future alerts with trips specified](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213994843939436?focus=true)

I'm a bit cautious of this change, but looking at the sequence of #397, #414, and #419, I'm pretty sure the current alert filtering is doing some redundant checks within both the scheduler filter and the significance check.

First, in #397, the alert filter was added in the scheduler, and `significance` was just doing the now `intrinsic_significance` checks. It also mixes up the params to `DateTime.diff`, which results in all future alerts getting through the filter no matter how far in the future they become active, and filters out any alerts that became active > 24 hours in the past. It also does the `all_clear?` check just by checking the presence of the closed timestamp.

Then, #414 added the max significance checks and the `active_soon?` helper (with the params in the right order), which makes the timeframe check in the scheduler redundant, because any alert that doesn't have an active/upcoming/past period will have a nil significance.

Then #419 adds the `Alert.all_clear?` function, which checks that all the active periods are in the past, but ignores the presence of the closed timestamp.

So in these changes, I added the closed timestamp check to `Alert.all_clear?`, it's only used in `AlertSummary.all_clear_summary` and `Alert.significance`, and I'm pretty sure both of them should be taking the closed timestamp into account. Then, since `Alert.significance` is checking both that the alert is either active, upcoming, or all clear, it should no longer be necessary to do those checks within the scheduler, and we can just check that the significance is non-nil.

We also recently added the check on [engine.ex#L169](https://github.com/mbta/mobile_app_backend/blob/main/lib/mobile_app_backend/notifications/engine.ex#L169) which only creates all clears when `alert.closed_timestamp == alert.last_push_notification_timestamp`. Would it also make sense to include that in `Alert.all_clear?`, then switch the engine check to also use that? Maybe the active periods being in the past don't matter in any of these cases and we can just use the closed timestamp being non-null and equal to the last push timestamp?

I'm not actually sure how this would prevent reminders, since negative times are still less than 24, and this change should make future alerts notify less frequently (but more correctly). Alerts over 24 hours in the future should also have still not been sent, because they would have a nil significance, so this would have only erroneously prevented alerts that started over 24 hours in the past from being sent. That said, I'm seeing trip specific notifications for alerts with future active periods now, so either this did fix it, or there was something else going wrong with the testing when this was reported.

## Testing

I added a few new tests to ensure that future alerts and all clears are being filtered properly, and did some manual testing with all clears and trip specific suspensions in dev.

### DIFF
--- a/lib/mbta_v3_api/alert.ex
+++ b/lib/mbta_v3_api/alert.ex
@@ -220,10 +220,11 @@ defmodule MBTAV3API.Alert do
 
   @spec all_clear?(t(), DateTime.t()) :: boolean()
   def all_clear?(alert, at_time) do
-    Enum.all?(
-      alert.active_period,
-      &(not is_nil(&1.end) and DateTime.before?(&1.end, at_time))
-    )
+    not is_nil(alert.closed_timestamp) and
+      Enum.all?(
+        alert.active_period,
+        &(not is_nil(&1.end) and DateTime.before?(&1.end, at_time))
+      )
   end
 
   @spec parse!(JsonApi.Item.t()) :: t()
--- a/lib/mbta_v3_api/alert.ex
+++ b/lib/mbta_v3_api/alert.ex
@@ -218,13 +218,10 @@ defmodule MBTAV3API.Alert do
     next_period(alert, now) != nil
   end
 
-  @spec all_clear?(t(), DateTime.t()) :: boolean()
-  def all_clear?(alert, at_time) do
+  @spec all_clear?(t()) :: boolean()
+  def all_clear?(alert) do
     not is_nil(alert.closed_timestamp) and
-      Enum.all?(
-        alert.active_period,
-        &(not is_nil(&1.end) and DateTime.before?(&1.end, at_time))
-      )
+      alert.closed_timestamp == alert.last_push_notification_timestamp
   end
 
   @spec parse!(JsonApi.Item.t()) :: t()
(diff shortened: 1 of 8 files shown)


lib/mbta_v3_api/alert.ex
```diff
@@ -220,10 +220,11 @@ defmodule MBTAV3API.Alert do
...
   def all_clear?(alert, at_time) do
-    Enum.all?(
-      alert.active_period,
-      &(not is_nil(&1.end) and DateTime.before?(&1.end, at_time))
-    )
+    not is_nil(alert.closed_timestamp) and
+      Enum.all?(
+        alert.active_period,
+        &(not is_nil(&1.end) and DateTime.before?(&1.end, at_time))
+      )
```
KaylaBrady: > So in these changes, I added the closed timestamp check to Alert.all_clear?, it's only used in AlertSummary.all_clear_summary and Alert.significance, and I'm pretty sure both of them should be taking the closed timestamp into account. Then, since Alert.significance is checking both that the alert is either active, upcoming, or all clear, it should no longer be necessary to do those checks within the scheduler, and we can just check that the significance is non-nil.

> We also recently added the check on [engine.ex#L169](https://github.com/mbta/mobile_app_backend/blob/main/lib/mobile_app_backend/notifications/engine.ex#L169) which only creates all clears when alert.closed_timestamp == alert.last_push_notification_timestamp. Would it also make sense to include that in Alert.all_clear?, then switch the engine check to also use that? Maybe the active periods being in the past don't matter in any of these cases and we can just use the closed timestamp being non-null and equal to the last push timestamp?

I think it would make sense to put `last_push_notification_timestamp` into this function and use it universally. I also do think we should remove the active period check so that we are matching the definition of all clear as provided by alerts manager. 

On the topic of Alert.significance, I personally find the concept of "max_significance" relative to time a bit confusing and kind of prefer keeping the significance separate from the time-based relevance. The only place we are using `significance` with a passed time is in scheduler, and that is only to check if significance != nil. So it feels safe to me to break that into two checks, or into a new named check (something like `is_visible_for_notifications` with a better name)

EmmaSimon: The time based significance is still necessary to downgrade major into secondary if they're upcoming, so we can't just use a boolean check for that, and need to do some kind of min check or explicit significance downgrade somewhere

KaylaBrady: Is that downgrade from major to secondary being used on the backend though? I see the only caller to `Alert.significance` that passes the time as the one in scheduler. I see where major => secondary is done on the frontend, but not sure the concept is used elsewhere on the backend

EmmaSimon: Yeah, I don't think it is currently, but I'm still a bit nervous about changing the behavior across the backend and frontend in case we ever start using the significance on the backend somehow. But I guess removing this would force us to reconsider it if/when we decide we want to do that.

KaylaBrady: Part of my motivation is I'm considering diverging it further on the frontend [here](https://github.com/mbta/mobile_app/pull/1659/changes#diff-fdba8dd30f71d7b10db808e55052c8efdc29e198b4e8d8af09fb3bfc004d787eR59). I don't totally love that approach though and am thinking of trying something else out... basically I'm worried about stuffing too much logic into significance when the logic applied on top of inherent significance is kind of case-by-case based on where it is used. 

Putting that aside though, I think my comment about changing significance can be considered non-blocking, and only the piece about `all_clear?` as blocking.

EmmaSimon: Yeah, that makes sense. I gave removing the time based significance a shot, and I'm okay with what I ended up with. I think this is clearer even if it doesn't match the frontend exactly.

KaylaBrady (approved):
🙌

PR #461, opened 2026-04-16
feat: read and expand schedules’ added routes

boringcactus:
### Summary

_Ticket:_ [Ensure multi-route schedules and predictions are correctly displayed](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1213913595242981?focus=true)

Reads https://github.com/mbta/api/pull/992 and turns a single schedule with added routes into multiple schedules. Creating multiple copies of the schedule in the backend means that we can treat schedules and predictions the same way in the frontend.


KaylaBrady (approved):
Great!

PR #463, opened 2026-04-17
fix(Notifications.Scheduler): rescue exceptions, and fix one exception happening now

KaylaBrady:
### Summary

_Ticket:_ [Prevent scheduler crashes](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214104746531461?focus=true)

What is this PR for?
This PR fixes the intended exception catching in the scheduler module by changing `catch` to `rescue`. This is a nitty difference I didn't think twice about since `catch` is the conventional word elsewhere! [From the docs](https://hexdocs.pm/elixir/try-catch-and-rescue.html#errors), rescue is for typical exceptions that are raised, and catch is for places you intentionally `throw` a value, which should be done ~never. "using try/catch is already uncommon and using it to catch exits is even rarer."

This also fixes the underlying GL issue thanks to help from @EmmaSimon and @boringcactus ([slack thread](https://mbta.slack.com/archives/C062QNAJZ2M/p1776451100556429)) and adds a couple more typespecs along the way.

Testing
* Ran without the GL fix to make sure exception catching & logging works as expected
image

* Added unit test for exception catching
* Re-ran with the GL fix to make sure Scheduler ran as expected

### DIFF
--- a/lib/mobile_app_backend/alerts/alert_summary.ex
+++ b/lib/mobile_app_backend/alerts/alert_summary.ex
@@ -209,11 +209,17 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
       patterns
       |> Enum.group_by(& &1.direction_id)
       |> Map.new(fn {direction_id, direction_patterns} ->
+        maybe_typical_patterns = Enum.find(direction_patterns, &(&1.typicality == :typical))
+
         stop_list =
-          get_stop_list_for_pattern(
-            Enum.find(direction_patterns, &(&1.typicality == :typical)),
-            global
-          )
+          if is_nil(maybe_typical_patterns) do
+            nil
+          else
+            get_stop_list_for_pattern(
+              maybe_typical_patterns,
+              global
+            )
+          end
 
         {direction_id, stop_list}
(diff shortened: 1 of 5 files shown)


lib/mobile_app_backend/alerts/alert_summary.ex
```diff
@@ -199,15 +202,24 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
...
     end
 
+    @spec get_typical_stop_list_by_direction([RoutePattern.t()], GlobalDataCache.data()) :: %{
+            integer() => [Stop.id()] | nil
+          }
     defp get_typical_stop_list_by_direction(patterns, global) do
       patterns
       |> Enum.group_by(& &1.direction_id)
       |> Map.new(fn {direction_id, direction_patterns} ->
+        maybe_typical_patterns = Enum.find(direction_patterns, &(&1.typicality == :typical))
```
boringcactus: Since this is only ever one pattern, it should probably have a singular name.

boringcactus (approved):

PR #464, opened 2026-04-17
feat: store user locale for fallback text

boringcactus:
### Summary

_Ticket:_ [Store top supported language for notifications fallback](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214048781923336?focus=true)

Pairs with https://github.com/mbta/mobile_app/pull/1681.

### DIFF
--- a/lib/mobile_app_backend/notifications/write_payload.ex
+++ b/lib/mobile_app_backend/notifications/write_payload.ex
@@ -1,5 +1,4 @@
 defmodule MobileAppBackend.Notifications.WritePayload do
-  require Util
   alias Ecto.Changeset
   alias MobileAppBackend.Notifications
   alias MobileAppBackend.User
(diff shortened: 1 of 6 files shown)


lib/mobile_app_backend/notifications/write_payload.ex
```diff
@@ -99,19 +100,28 @@ defmodule MobileAppBackend.Notifications.WritePayload do
...
-  def parse!(%{"fcm_token" => fcm_token, "subscriptions" => subscriptions}) do
+  def parse!(%{"fcm_token" => fcm_token, "subscriptions" => subscriptions} = payload) do
     %__MODULE__{
       fcm_token: fcm_token,
-      subscriptions: MapSet.new(subscriptions, &Subscription.parse!/1)
+      subscriptions: MapSet.new(subscriptions, &Subscription.parse!/1),
+      locale:
+        case Map.fetch(payload, "locale") do
+          {:ok, locale} when Util.is_known_locale(locale) -> locale
+          _ -> nil
```
KaylaBrady: question: Why do we need to check if it is a known locale to save it? It is unlikely that we add a language anytime soon, but it feels safer to do an is_known_locale check at the time of fetching the appropriate translated content (which we don't have a plan to introduce near term) rather than at the time of storage, in case we ever update the supported languages on the frontend before the backend

boringcactus: I guess the design of https://github.com/mbta/mobile_app/pull/1681 makes it very difficult to accidentally introduce locales that aren’t actually valid, so we can assume that any new locale that shows up is valid.

KaylaBrady (approved):

PR #470, opened 2026-04-23
feat: Update trip summary behavior

EmmaSimon:
### Summary

_Ticket:_ [Notifications QA | Trip specific suspension and shuttle alert summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214072175090608?focus=true)
Frontend PR: https://github.com/mbta/mobile_app/pull/1685

Fix a few edge cases with trip specific shuttle, suspension, and stop closure legibility.
I added the `ThisTrip` state, even though we don't use it anywhere on the backend, because it seemed weird to have this not match the frontend, but maybe it's worth removing it so that we have no chance of accidentally sending it to the frontend.

Added a few unit tests, but most cases were already covered by existing ones.

IMG_0050
IMG_0051
IMG_0052
IMG_0054
IMG_0053

### DIFF
--- a/lib/mobile_app_backend/alerts/alert_summary.ex
+++ b/lib/mobile_app_backend/alerts/alert_summary.ex
@@ -228,27 +228,39 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
 
   defmodule Location do
     defmodule DirectionToStop do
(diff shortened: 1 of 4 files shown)


KaylaBrady (approved):
Great! I think good to add `ThisTrip` even if unused at this point with the upcoming ticket for considering consolidation.

PR #476, opened 2026-04-30
fix: Only send alerts with push timestamps

EmmaSimon:
### Summary

_Ticket:_ [Notifications | Fix: users receive notifications for alerts created with notification checkbox unchecked](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214312229016953?focus=true)

I think this should cover all cases of only notifying users if the checkbox updates the push timestamp. Since nil timestamps are filtered out, that ensures that a notification or reminder will never be sent unless the notification checkbox was ticket on creation or a subsequent update, and since now the push timestamp is the only time that can ever be saved with the notification record, new notifications can't be sent until the timestamp is updated, and all clears have their separate check to make sure the close timestamp is the same as the push timestamp.

### DIFF
--- a/test/mobile_app_backend/notifications/engine_test.exs
+++ b/test/mobile_app_backend/notifications/engine_test.exs
@@ -22,8 +22,7 @@ defmodule MobileAppBackend.Notifications.EngineTest do
       build(:alert,
         active_period: [%Alert.ActivePeriod{start: DateTime.from_unix!(0), end: nil}],
         effect: :suspension,
-        informed_entity: [%Alert.InformedEntity{activities: [:board], route: "Green-D"}],
-        last_push_notification_timestamp: now |> DateTime.add(-2)
+        informed_entity: [%Alert.InformedEntity{activities: [:board], route: "Green-D"}]
       )
 
     subscription =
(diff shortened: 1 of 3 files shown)


test/mobile_app_backend/notifications/engine_test.exs
```diff
@@ -22,7 +22,8 @@ defmodule MobileAppBackend.Notifications.EngineTest do
       build(:alert,
         active_period: [%Alert.ActivePeriod{start: DateTime.from_unix!(0), end: nil}],
         effect: :suspension,
-        informed_entity: [%Alert.InformedEntity{activities: [:board], route: "Green-D"}]
+        informed_entity: [%Alert.InformedEntity{activities: [:board], route: "Green-D"}],
+        last_push_notification_timestamp: now |> DateTime.add(-2)
```
KaylaBrady: suggestion(non-blocking): maybe worth adding a default to the factory

KaylaBrady (approved):
Appreciate the approach of filtering at the highest level!

test/mobile_app_backend/notifications/engine_test.exs
EmmaSimon: Good call!

PR #482, opened 2026-05-04
fix: Include active alert in next hour as upcoming

EmmaSimon:
### Summary

_Ticket:_ [Notifications QA | No alert sent for cancellation of future trip](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1214429661481092?focus=true)

Found it 😅 

`DateTime.diff(ap_start, now, :hour)` truncates any fraction, so alerts between now and now + 1hr will return 0, which `Alert.next_period` considered to be happening now, so it wasn't included in currently active or soon active.


KaylaBrady (approved):
Good find!

PR #483, opened 2026-05-05
feat: port PushNotificationPayload.getDeepLinkState to the backend

boringcactus:
### Summary

_Ticket:_ [[fix notification deep linking and icon and stuff]](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214547310290612)


lib/mobile_app_backend/notifications/deliverer.ex
```diff
@@ -35,12 +36,30 @@ defmodule MobileAppBackend.Notifications.Deliverer do
...
       message: %FCM.Model.Message{
         notification: %{
           title: title,
           body: body
         },
+        data: %{deep_link_path: deep_link_path},
+        android: %FCM.Model.AndroidConfig{
+          notification: %FCM.Model.AndroidNotification{
+            # TODO why does this not cause sound to play
+            sound: "default",
```
boringcactus, line 56: According to [the docs](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#AndroidNotification), this Should™ work, but it appears not to, and I can’t tell if that’s some weird artifact of my setup or something that’s actually wrong and needs fixing.

lib/mobile_app_backend/notifications/scheduler.ex
```diff
@@ -183,6 +183,41 @@ defmodule MobileAppBackend.Notifications.Scheduler do
...
+      |> case do
+        [%{direction: direction}] -> "/d/#{direction}"
+        _ -> ""
+      end
+
+    if stop_piece != "" do
+      stop_piece <> route_piece <> direction_piece
+    else
+      "/a/#{alert_id}" <> route_piece <> stop_piece
+    end
```
boringcactus, line 218: I’m not including the actual host here, because I’m not quite sure if the backend knows its own domain name, and because it’s an unnecessary use of payload space (although testing that the frontend will accept a bare URL fragment is a little fiddly).

lib/mobile_app_backend/notifications/deliverer.ex
KaylaBrady: Confirmed sound plays for me!

lib/mobile_app_backend/notifications/deliverer.ex
```diff
@@ -35,12 +36,30 @@ defmodule MobileAppBackend.Notifications.Deliverer do
...
+            tag: tag
```
KaylaBrady, line 57: I'm assuming this is where we'd want to specify the icon too if we want to specify it for android. The icon already looks right on my device. I think we can consider that non-blocking for this PR, and I'll look into it more as a follow-up.

KaylaBrady (approved):

lib/mobile_app_backend/notifications/deliverer.ex
KaylaBrady: Oh from the [doc](https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages#AndroidNotification)

> The notification's icon. Sets the notification icon to myicon for drawable resource myicon. If you don't send this key in the request, FCM displays the launcher icon specified in your app manifest.

So maybe this has to do more with an issue with the launcher icon? Will look closer, but now even more willing to call this PR done.

PR #500, opened 2026-05-07
chore: build route branching diagram for non-branching routes too

boringcactus:
### Summary

_Ticket:_ none

This should make the diffs easier to follow when a route changes from non-branching to branching or vice versa.


KaylaBrady (approved):
Much appreciated!!

PR #514, opened 2026-05-18
feat: distinguish notification from update in type

boringcactus:
### Summary

_Ticket:_ [Notifications | Click analytics](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214587659492421)

I split this out from #513 since I wasn’t sure how I wanted to do this, but I think this approach makes sense and isn’t too complicated.

### DIFF
--- a/lib/mobile_app_backend/notifications/delivered_notification.ex
+++ b/lib/mobile_app_backend/notifications/delivered_notification.ex
@@ -5,8 +5,7 @@ defmodule MobileAppBackend.Notifications.DeliveredNotification do
   alias MobileAppBackend.Repo
   alias MobileAppBackend.User
 
-  @type provisional_type :: :reminder | {:notification_or_update, DateTime.t()} | :all_clear
-  @type final_type :: :reminder | {:notification | :update, DateTime.t()} | :all_clear
+  @type type :: :reminder | {:notification | :update, DateTime.t()} | :all_clear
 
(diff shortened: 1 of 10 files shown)


lib/mobile_app_backend/notifications/delivered_notification.ex
```diff
@@ -15,14 +16,34 @@ defmodule MobileAppBackend.Notifications.DeliveredNotification do
...
     )
 
     timestamps(type: :utc_datetime)
   end
 
-  @spec can_send?(User.id(), Alert.id(), type()) :: boolean()
+  @spec finalize_type(User.id(), Alert.id(), provisional_type()) :: final_type()
+  def finalize_type(user_id, alert_id, type)
+
+  def finalize_type(user_id, alert_id, {:notification_or_update, upstream_timestamp}) do
```
KaylaBrady: question: Is there a nice spot to do this earlier in the flow to be able to avoid the provisional type entirely?

boringcactus: We can check this in the engine, but that means all the engine tests need subscriptions with a non-nil user ID, which is why I didn’t do it at first. In hindsight, maybe having slightly more complicated tests is better than having a slightly more complicated implementation, especially since the user ID doesn’t have to actually exist.

KaylaBrady (approved):

PR #523, opened 2026-06-01
feat: Include summary entities in v3 alert stream

EmmaSimon:
### Summary

_Ticket:_ [Alert Summary Consolidation | Initial setup of new Alerts V3 Channel (only give current state on join)](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214749715355765?focus=true)

This change adds a new field to alert objects which contains summary strings for each combination or route, stop, trip, and direction that the alert effects. The new entities mirror the structure of informed entities, with each containing nillable IDs for the route, stop, trip, and direction that they apply to (with nils as wildcards).

I opted to include the entities directly in the alert object so that we don't have to mess around with prop drilling another object on the frontend.

This PR doesn't include:
* Caching the generated summaries
* Any frontend vs backend summary differences ("this trip" vs "11:00 train from X")
* Diffing payloads with only new or updated alerts

Examples:
Screenshot 2026-06-01 at 3 14 46 PM
Screenshot 2026-06-01 at 3 15 09 PM
Screenshot 2026-06-01 at 3 16 12 PM
Screenshot 2026-06-01 at 3 21 33 PM
Screenshot 2026-06-01 at 3 20 23 PM

### DIFF
--- a/lib/mbta_v3_api/alert.ex
+++ b/lib/mbta_v3_api/alert.ex
@@ -7,6 +7,7 @@ defmodule MBTAV3API.Alert do
   alias MBTAV3API.RoutePattern
   alias MBTAV3API.Stop
   alias MBTAV3API.Trip
+  alias MobileAppBackend.Alerts.SummaryEntity
 
   @type t :: %__MODULE__{
           id: String.t(),
@@ -22,6 +23,7 @@ defmodule MBTAV3API.Alert do
           last_push_notification_timestamp: DateTime.t() | nil,
           lifecycle: lifecycle(),
           severity: integer(),
+          summaries: [SummaryEntity.t()] | nil,
           updated_at: DateTime.t()
         }
 
@@ -158,6 +160,7 @@ defmodule MBTAV3API.Alert do
     :last_push_notification_timestamp,
     :lifecycle,
     :severity,
+    :summaries,
     :updated_at
   ]
 
@@ -176,6 +179,7 @@ defmodule MBTAV3API.Alert do
       :last_push_notification_timestamp,
       :lifecycle,
       :severity,
+      :summaries,
       :updated_at
     ]
   end
--- a/lib/mbta_v3_api/alert.ex
+++ b/lib/mbta_v3_api/alert.ex
@@ -7,7 +7,6 @@ defmodule MBTAV3API.Alert do
   alias MBTAV3API.RoutePattern
   alias MBTAV3API.Stop
   alias MBTAV3API.Trip
-  alias MobileAppBackend.Alerts.SummaryEntity
 
   @type t :: %__MODULE__{
           id: String.t(),
@@ -23,7 +22,6 @@ defmodule MBTAV3API.Alert do
           last_push_notification_timestamp: DateTime.t() | nil,
           lifecycle: lifecycle(),
           severity: integer(),
-          summaries: [SummaryEntity.t()] | nil,
           updated_at: DateTime.t()
         }
 
@@ -160,7 +158,6 @@ defmodule MBTAV3API.Alert do
     :last_push_notification_timestamp,
     :lifecycle,
     :severity,
-    :summaries,
     :updated_at
   ]
 
@@ -179,7 +176,6 @@ defmodule MBTAV3API.Alert do
       :last_push_notification_timestamp,
       :lifecycle,
       :severity,
-      :summaries,
       :updated_at
     ]
   end
--- a/lib/mbta_v3_api/route_pattern.ex
+++ b/lib/mbta_v3_api/route_pattern.ex
@@ -1,5 +1,6 @@
 defmodule MBTAV3API.RoutePattern do
   alias MBTAV3API.Trip
+  alias MobileAppBackend.GlobalDataCache
   use MBTAV3API.JsonApi.Object
   require Util
 
@@ -95,7 +96,7 @@ defmodule MBTAV3API.RoutePattern do
 
   @spec canonical_or_most_typical([t()]) :: [t()]
   # Filter the list of route patterns to include only the canonical or most typical patterns
-  defp canonical_or_most_typical(route_patterns) do
+  def canonical_or_most_typical(route_patterns) do
     canonical_patterns = Enum.filter(route_patterns, & &1.canonical)
 
     case canonical_patterns do
@@ -126,5 +127,32 @@ defmodule MBTAV3API.RoutePattern do
     route_patterns
     |> Enum.map(& &1.route)
     |> Map.new(&{&1.id, &1})
+  end
+
+  @spec get_relevant_patterns(
+          String.t() | nil,
+          String.t() | nil,
+          integer() | nil,
+          GlobalDataCache.data()
+        ) :: [t()]
+  def get_relevant_patterns(route_id, stop_id, direction_id, global_data) do
+    global_data.route_patterns
+    |> Stream.map(fn {_, pattern} -> pattern end)
+    |> Enum.filter(fn pattern ->
+      (route_id == nil or pattern.route_id == route_id or
+         global_data.routes[pattern.route_id].line_id == route_id) and
+        (direction_id == nil or pattern.direction_id == direction_id) and
+        (stop_id == nil or
+           Enum.any?(
+             with trip_id when is_binary(trip_id) <- pattern.representative_trip_id,
+                  %Trip{} = trip <- global_data.trips[trip_id] do
+               trip.stop_ids
+             else
+               _ -> []
+             end,
+             &(&1 == stop_id or
+                 &1 in global_data.stops[stop_id].child_stop_ids)
+           ))
+    end)
   end
 end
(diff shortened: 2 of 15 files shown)


lib/mbta_v3_api/alert.ex
```diff
@@ -22,6 +23,7 @@ defmodule MBTAV3API.Alert do
           last_push_notification_timestamp: DateTime.t() | nil,
           lifecycle: lifecycle(),
           severity: integer(),
+          summaries: [SummaryEntity.t()] | nil,
```
KaylaBrady: suggestion: Pretty minor, but I think it would be worth creating a new alert object, maybe in `/mobile_app_backend/alerts` that has the new field to keep this representation as purely what we get from the API

lib/mbta_v3_api/route_pattern.ex
```diff
@@ -127,4 +128,31 @@ defmodule MBTAV3API.RoutePattern do
...
+  @spec get_relevant_patterns(
+          String.t() | nil,
+          String.t() | nil,
+          integer() | nil,
+          GlobalDataCache.data()
+        ) :: [t()]
+  def get_relevant_patterns(route_id, stop_id, direction_id, global_data) do
+    global_data.route_patterns
+    |> Stream.map(fn {_, pattern} -> pattern end)
+    |> Enum.filter(fn pattern ->
```
KaylaBrady, line 141: suggestion(non-blocking): The long string of boolean operators makes this a little hard to read. Maybe breaking out some named variables or helper functions would make it clearer

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -0,0 +1,295 @@
...
+  @spec build_all([Alert.t()], DateTime.t(), GlobalDataCache.data()) :: %{
+          String.t() => [SummaryEntity.t()]
+        }
+  def build_all(alerts, at_time, global) do
+    Map.new(Enum.map(alerts, fn alert -> {alert.id, build_for_alert(alert, at_time, global)} end))
+  end
+
+  @spec build_all([Alert.t()]) :: %{String.t() => [SummaryEntity.t()]}
+  def build_all(alerts) do
+    at_time = DateTime.now!("America/New_York")
```
KaylaBrady: suggestion: passing a time param that defaults to now can help with testability

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -0,0 +1,295 @@
...
+
+    build_all(alerts, at_time, global)
+  end
+
+  @spec build_for_alert(Alert.t(), DateTime.t(), GlobalDataCache.data()) :: [SummaryEntity.t()]
+  defp build_for_alert(alert, at_time, global) do
+    # Fetch schedules once for the whole alert for any trips included in the informed entities
+    {schedules, trips} = fetch_schedules_for_alert(alert)
+
+    # The global stops don't include all child stops, so we need to fetch them separately
```
KaylaBrady, line 51: question(non-blocking): Why doesn't it include all child stops? Would it be worth just including them in global?

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -0,0 +1,295 @@
...
+      end)
+    end)
+    |> dedup_summaries()
+  end
+
+  # Get the route, stop, trip, and direction combinations relevant to the alert based on its informed_entity list
+  @spec relevant_combinations(Alert.t(), %{String.t() => Stop.t()}, GlobalDataCache.data()) :: [
+          {String.t() | nil, String.t() | nil, String.t() | nil, 0 | 1 | nil}
+        ]
+  defp relevant_combinations(alert, stops, global) do
```
KaylaBrady: suggestion(non-blocking): May be worth exposing this and/or other helpers for testability

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -0,0 +1,295 @@
...
+    # expanded to both directions
+    Enum.reject(combinations, fn combination ->
+      redundant_combination?(combination, 0, combinations) and
+        redundant_combination?(combination, 1, combinations) and
+        redundant_combination?(combination, 2, combinations)
+    end)
+  end
+
+  @spec combination_from_entity(Alert.InformedEntity.t(), GlobalDataCache.data()) :: [
+          {String.t() | nil, String.t() | nil, String.t() | nil, 0 | 1 | nil}
```
KaylaBrady, line 147: suggestion: I think it might be worth making a struct for this for readability so that we don't have to use `elem`

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -0,0 +1,295 @@
...
+  defp resolve_stop_id(stop_id, trip_id, patterns, trips, stops, global) do
+    stop_id ||
+      stop_parent_id(
+        case trip_id do
+          nil ->
+            RoutePattern.canonical_or_most_typical(patterns)
+            |> List.first()
+            |> case do
+              %RoutePattern{representative_trip_id: representative_trip_id} ->
+                global.trips[representative_trip_id].stop_ids |> List.first()
```
KaylaBrady, line 232: question: Why use the first stop from the representative trip? May be worth a function doc

lib/mobile_app_backend_web/channels/alerts_channel.ex
```diff
@@ -10,7 +10,12 @@ defmodule MobileAppBackendWeb.AlertsChannel do
         MobileAppBackend.Alerts.PubSub
       )
 
-    data = pubsub_module.subscribe(legacy_compatibility: true)
+    data =
+      pubsub_module.subscribe(
+        legacy_compatibility: true,
+        include_summaries: false
```
KaylaBrady: I like the way you used a flag for this!

KaylaBrady (commented):
This is great! I really like the approach you're taking of using the informed entities wildcard matching. Have you had a chance to measure how much this increases the size of the join payload?

lib/mobile_app_backend/alerts/summary_entity_builder.ex
EmmaSimon: I included the other `build_all` above which accepts the time and global, I figured I'd use that for testing

lib/mobile_app_backend/alerts/summary_entity_builder.ex
EmmaSimon: That's what I did at first, but the frontend expects only `location_type` 0 and 1 in global, and there are some missing fields on the other types that started breaking things.

lib/mobile_app_backend/alerts/summary_entity_builder.ex
KaylaBrady: Oh lol yep I scanned ed right over that!

lib/mobile_app_backend/alerts/summary_entity_builder.ex
EmmaSimon: Yeah, I was a bit unsure if this would definitely cover all possible cases, but `summarizing` needs _some_ stop ID, and I couldn't think of a better way to handle it

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -221,10 +221,15 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
...
+    # If a stop ID is provided in the informed entities, we can return it directly,
+    # but if not, we need to determine a relevant stop from the pattern or trip
     stop_id ||
       stop_parent_id(
         case trip_id do
           nil ->
+            # We need to provide a specific stop ID to AlertSummary.summarizing, but if the informed entity
+            # only specifies a route ID without a stop or trip, the only way for us to get a relevant stop ID
+            # is to pull one from a pattern's representative trip, since the summary shouldn't be different across stops
             RoutePattern.canonical_or_most_typical(patterns)
```
KaylaBrady, line 233: thought(non-blocking): I wonder if making stop ID nullable for `summarizing` or having different entrypoints for `summarizing` would help. Maybe worth thinking about later on

lib/mobile_app_backend/application.ex
```diff
@@ -5,6 +5,9 @@ defmodule MobileAppBackend.Application do
 
   use Application
 
+  @default_locale Application.compile_env!(:mobile_app_backend, :default_locale_code)
+  def default_locale, do: @default_locale
```
KaylaBrady, line 9: Ooh this is nice!

KaylaBrady (approved):
This is great!!

lib/mobile_app_backend/alerts/summary_entity_builder.ex
EmmaSimon: Yeah I think that would be worth looking into, I can make a ticket. Edit: [done](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1215452396175523?focus=true)

PR #533, opened 2026-06-09
fix(Notifications.Window): Handle daylight savings

KaylaBrady:
### Summary

_Ticket:_ [Investigate error scheduling notifications](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215500480088770?focus=true)

The cause of the failure was a recurring alert that had active periods out into 2027. `Window.next_overlap` was creating date times combining the active period dates & the user's window time, which could result in invalid dates due to DST. This now safely handles those scenarios by using a new function which always chooses the later valid time around the DST boundary.

While adding this new util, I opted to break out a separate `Utils.DateTime` file for organization.

### Testing

* Added tests for the new function


EmmaSimon (approved):
Good call splitting out the separate module

PR #538, opened 2026-06-15
feat: generate alert summary table

boringcactus:
### Summary

_Ticket:_ [Generate alert summary table](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215434236058781?focus=true)

I may have overthought this DSL a little bit, but c’mon, look how readable those scenario definitions are!

### Testing

Checked that the resulting `.csv` looks defensible.

### DIFF
--- a/lib/mix/tasks/generate_alert_summary_table.ex
+++ b/lib/mix/tasks/generate_alert_summary_table.ex
@@ -143,18 +143,22 @@ if Mix.env() == :test do
     end
 
     defp here_and_now do
-      scenario("Here + Now", ~N[2026-06-10 12:23:00])
+      wednesday_noon = ~N[2026-06-10 12:00:00]
+
+      scenario("Here + Now", wednesday_noon)
       |> subscription(route: "Orange", stop: "place-rugg", direction: 0)
       |> alert(cause: :maintenance, effect: :suspension)
-      |> active_period(~N[2026-06-08 12:00:00], ~N[2026-06-12 12:00:00])
+      |> active_period({-2, :days}, {2, :days})
       |> informed_entity(stops: ["place-rugg", "place-rcmnl", "place-jaksn"])
     end
 
     defp here_and_now_os_notification do
-      scenario("Here + Now (OS Notification)", ~N[2026-06-12 12:23:00])
+      friday_noon = ~N[2026-06-12 12:00:00]
+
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/generate_alert_summary_table.ex
```diff
@@ -0,0 +1,720 @@
...
+    defp here_and_now_os_notification_multiple_routes do
+      scenario("Here + Now (OS Notification, Multiple Routes)", ~N[2026-06-10 12:23:00])
+      |> subscription(route: "Orange", stop: "place-north", direction: 0)
+      |> subscription(route: "line-Green", stop: "place-north", direction: 0)
+      |> alert(effect: :station_closure)
+      |> active_period(~N[2026-06-08 12:00:00], ~N[2026-06-12 12:00:00])
+      |> informed_entity(routes: ["Orange", "Green-D", "Green-E"], stops: ["place-north"])
+    end
+
+    defp here_and_now_rl_shuttle do
```
KaylaBrady, line 190: suggestion(non-blocking): I think it would help readability to define a few constants for the times to reference in each scenario since the individual times don't matter as much as the relationships between them.

lib/mix/tasks/generate_alert_summary_table.ex
```diff
@@ -0,0 +1,720 @@
...
+        cr_trip_suspended_here(),
+        cr_trip_suspended_downstream(),
+        cr_trip_shuttle_here(),
+        cr_trip_shuttle_downstream(),
+        cr_trip_station_bypass_here(),
+        cr_trip_station_bypass_downstream()
+      ]
+    end
+
+    defp here_and_now do
```
KaylaBrady, line 145: I do like how straightforward defining each scenario is! This is a bit different from what I had initially imagined (ex: defining an input for routes, effect, active period, etc. and iterating), but considering that would produce so many implausible scenarios, I like this better.

KaylaBrady (approved):

lib/mix/tasks/generate_alert_summary_table.ex
boringcactus: How’s 6f4ad7c?

KaylaBrady: LGTM!

PR #528, opened 2026-06-05
fix: Address zizmor analysis and add CI step

EmmaSimon:
### Summary

_Ticket:_ [Run Zizmor & fix any GHA vulnerabilities](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214938277834907?focus=true)

Address zizmor results, this passes locally when run in regular mode.

### Testing

CI actions pass, tbd on deploys

PR #545, opened 2026-06-16
build: fix deploy workflows after zizmor

boringcactus:
### Summary

_Ticket:_ none

All our deploys appear to be broken after #528 because the deploy workflow is called with insufficient permissions. This fixes it.

### Testing

Deployed this PR to dev-orange, so it definitely works.


EmmaSimon (approved):
Oops, good catch 🙏

PR #550, opened 2026-06-18
test: Flaky test roundup

KaylaBrady:
### Summary

_Ticket:_ [Flaky test roundup](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1214097786423207?focus=true)

I was not able to reproduce the specific failure mentioned in the test. But these changes make it so that we can reliably run tests with `--repeat-until-failure` so that we have an easier time chasing down flaky tests in the future. 

### Testing

* Ran locally with `--repeat-until-failure` on for 100 runs without failure *sometimes* (running for a higher number of repetitions I get memory errors). I still get misc errors including the ones described in the comments sometimes... but this is a start

### DIFF
--- a/test/mbta_v3_api/json_api/object_test.exs
+++ b/test/mbta_v3_api/json_api/object_test.exs
@@ -139,6 +139,11 @@ defmodule MBTAV3API.JsonApi.ObjectTest do
                    fn ->
                      Code.compile_quoted(bad_module)
                    end
+
+      on_exit(fn ->
+        :code.purge(BadModule)
+        :code.delete(BadModule)
+      end)
     end
 
     test "accepts renames" do
@@ -158,6 +163,11 @@ defmodule MBTAV3API.JsonApi.ObjectTest do
         end
 
       assert [{GoodModule, _}] = Code.compile_quoted(good_module)
+
+      on_exit(fn ->
+        :code.purge(GoodModule)
+        :code.delete(GoodModule)
+      end)
     end
   end
 
--- a/test/mbta_v3_api/stream/static_instance_test.exs
+++ b/test/mbta_v3_api/stream/static_instance_test.exs
@@ -91,7 +91,7 @@ defmodule MBTAV3API.Stream.StaticInstanceTest do
     end
 
     test "launches new instance if not already running" do
-      topic = "predictions:route:fake-route-that-won't-already-exist"
+      topic = "predictions:route:#{Uniq.UUID.uuid7()}"
       refute Stream.Registry.find_pid(topic)
       assert {:ok, _} = Stream.StaticInstance.subscribe(topic)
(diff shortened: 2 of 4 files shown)


test/mbta_v3_api/json_api/object_test.exs
```diff
@@ -139,6 +139,11 @@ defmodule MBTAV3API.JsonApi.ObjectTest do
                    fn ->
                      Code.compile_quoted(bad_module)
                    end
+
+      on_exit(fn ->
```
KaylaBrady, line 143: caused warning when running repeatedly b/c these kept getting re-created

test/mbta_v3_api/stream/static_instance_test.exs
```diff
@@ -91,7 +91,7 @@ defmodule MBTAV3API.Stream.StaticInstanceTest do
     end
 
     test "launches new instance if not already running" do
-      topic = "predictions:route:fake-route-that-won't-already-exist"
+      topic = "predictions:route:#{Uniq.UUID.uuid7()}"
```
KaylaBrady, line 94: This is kind of a hack for running repeatedly, otherwise on the re-run the topic would already have been joined and the test would fail. I'd rather have this test cleanup after itself, but I wasn't sure the best way to do it - killing the pid returned by `Stream.Registry.find_pid(topic)` didn't do it. 

I also tried using `start_link_supervised!(Stream.Supervisor)`, but this would require removing it from `MBTAV3API.Supervisor` and adding the start_link call in a bunch of other tests, which made me wonder if there is a better approach I'm missing.

test/mobile_app_backend/alerts/pub_sub_test.exs
```diff
@@ -1,16 +1,16 @@
...
+  alias MBTAV3API.JsonApi.Object
+  alias MBTAV3API.{Alert, Store, Stream}
   alias MobileAppBackend.Alerts.PubSub
   import Mox
   import Test.Support.Helpers
-  import Test.Support.Sigils
   import MobileAppBackend.Factory
 
   setup do
+    reassign_env(:mobile_app_backend, :alerts_broadcast_interval_ms, 10_000)
```
KaylaBrady, line 14: Occasionally getting test failures when running repeatedly that we are getting more calls than we expect (ex: `(Mox.UnexpectedCallError) expected VehiclesStoreMock.fetch/1 to be called 4 times but it has been called 5 times in process #PID<0.16044.0>`,). My hypothesis was that it was due to the test occasionally hitting the the 500ms broadcast interval.

I was also wondering if `:set_mox_from_context` could be causing some interference across test runs... but I couldn't find anything concrete with that. 

When running the tests repeatedly I do still occasionally hit these errors, but maybe less frequently?

EmmaSimon (approved):

PR #565, opened 2026-06-26
feat: Diff alerts:v3 channel updates

EmmaSimon:
### Summary

_Ticket:_ [Alert Summary Consolidation | Send diffs for alert changes](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215637601111265?focus=true)

This uses socket assigns to remember the hashes of the previously sent alerts, which are then compared against incoming alerts. All alerts are always sent on join, then subsequent messages are sent as a list of IDs to remove, and a map of alert data to update the client with.

### Testing

Added tests for the new behavior, also updated https://github.com/mbta/mobile_app/pull/1767 with the frontend changes and verified that updates were sent and received successfully

### DIFF
--- a/lib/mobile_app_backend_web/channels/alerts_channel.ex
+++ b/lib/mobile_app_backend_web/channels/alerts_channel.ex
@@ -1,8 +1,21 @@
 defmodule MobileAppBackendWeb.AlertsChannel do
   use MobileAppBackendWeb, :channel
 
+  defmodule AlertUpdate do
+    @type t :: %__MODULE__{
+            remove: [String.t()],
+            update: %{String.t() => Alert.t() | AlertWithSummaries.t()}
+          }
+    @derive Jason.Encoder
+    defstruct [:remove, :update]
+
+    def empty?(update) do
+      Enum.empty?(update.remove) and Enum.empty?(update.update)
+    end
+  end
+
   @impl true
-  def join("alerts" = topic, _payload, socket) do
+  def join("alerts", _payload, socket) do
     pubsub_module =
       Application.get_env(
         :mobile_app_backend,
@@ -10,12 +23,12 @@ defmodule MobileAppBackendWeb.AlertsChannel do
         MobileAppBackend.Alerts.PubSub
       )
 
-    data = pubsub_module.subscribe(get_opts(topic, socket))
+    data = pubsub_module.subscribe(get_opts(socket))
     {:ok, data, socket}
   end
 
   @impl true
-  def join("alerts:v2" = topic, _payload, socket) do
+  def join("alerts:v2", _payload, socket) do
     pubsub_module =
       Application.get_env(
         :mobile_app_backend,
@@ -23,12 +36,12 @@ defmodule MobileAppBackendWeb.AlertsChannel do
         MobileAppBackend.Alerts.PubSub
       )
 
-    data = pubsub_module.subscribe(get_opts(topic, socket))
+    data = pubsub_module.subscribe(get_opts(socket))
     {:ok, data, socket}
   end
 
   @impl true
-  def join("alerts:v3" = topic, _payload, socket) do
+  def join("alerts:v3", _payload, socket) do
     pubsub_module =
       Application.get_env(
         :mobile_app_backend,
@@ -36,12 +49,14 @@ defmodule MobileAppBackendWeb.AlertsChannel do
         MobileAppBackend.Alerts.PubSub
       )
 
-    data = pubsub_module.subscribe(get_opts(topic, socket))
-    {:ok, data, socket}
+    %{alerts: alerts} = pubsub_module.subscribe(get_opts(socket))
+    socket = assign(socket, :last_alerts, alert_hashes(alerts))
+
+    {:ok, %AlertUpdate{remove: [], update: alerts}, socket}
   end
 
-  defp get_opts(topic, socket) do
-    case topic do
+  defp get_opts(socket) do
+    case socket.topic do
       "alerts" ->
         [legacy_compatibility: true, include_summaries: false]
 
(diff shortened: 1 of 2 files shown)


lib/mobile_app_backend_web/channels/alerts_channel.ex
```diff
@@ -1,47 +1,62 @@
 defmodule MobileAppBackendWeb.AlertsChannel do
   use MobileAppBackendWeb, :channel
 
+  defmodule AlertUpdate do
+    @type t :: %__MODULE__{
+            remove: [String.t()],
+            update: %{String.t() => Alert.t() | AlertWithSummaries.t()}
```
KaylaBrady: question(non-blocking): wouldn't update always be an `AlertWithSummaries`?

lib/mobile_app_backend_web/channels/alerts_channel.ex
```diff
@@ -61,7 +76,72 @@ defmodule MobileAppBackendWeb.AlertsChannel do
...
+      "alerts:v3" ->
+        {:noreply, handle_info_v3({:new_alerts, data}, socket)}
+
+      _ ->
+        :ok = push(socket, "stream_data", data)
+        {:noreply, socket}
+    end
+  end
+
+  defp handle_info_v3({:new_alerts, %{alerts: alerts}}, socket) do
```
KaylaBrady, line 91: thought(non-blocking): I so wish I remembered this earlier, but the [generic pubsub](https://github.com/mbta/mobile_app_backend/blob/435df8ff3b957ae4ef082f3044de92c147680a04/lib/mobile_app_backend/pub_sub.ex#L31) already has functionality for looking at the last message sent vs the new one. Maybe that could be modified to do the diff calculation?

If we did want to do that, I don't think it needs to block this from being merged since the interface with the frontend wouldn't change.

lib/mobile_app_backend_web/channels/alerts_channel.ex
```diff
@@ -61,7 +76,72 @@ defmodule MobileAppBackendWeb.AlertsChannel do
...
+      }
+
+    if !AlertUpdate.empty?(response) do
+      :ok = push(socket, "stream_data", response)
+    end
+
+    assign(socket, :last_alerts, current_hashes)
+  end
+
+  @spec alert_hashes(%{String.t() => Alert.t()}) :: %{String.t() => String.t()}
```
KaylaBrady, line 120: thought(non-blocking): I wonder if instead of hashing the alerts, we could just use the `updated_at` field? Although now that I say that I realize that the summaries could be updated after the `updated_at` timestamp, thought we could separately store that timestamp in `AlertWithSummaries`... 

Again, I think no real need to change the implementation here, I was just thinking through if there is an easier way to detect a diff

KaylaBrady (approved):
Sweet! Spitballing a few different ideas which may not be practical but seemed worth mentioning

lib/mobile_app_backend_web/channels/alerts_channel.ex
EmmaSimon: Good point, yeah

lib/mobile_app_backend_web/channels/alerts_channel.ex
EmmaSimon: Yeah, I did consider using that, but the last dispatched table is keyed by the fetch keys (trip id for predictions, empty list for alerts), so I was thinking that we'd maybe instead need to use the pid for each subscriber as the key, and that seemed like it would have been too complex a refactor. But actually since alert data should always be the same for everyone, it would be safe to always send the same diff to everyone too? I was a little worried about taking an approach where we didn't look at each individual channel's last sent data, but I can't actually think of cases where it would cause problems with getting out of sync if we did do that.

lib/mobile_app_backend_web/channels/alerts_channel.ex
EmmaSimon: Ah, yeah that's a good idea, I didn't think of that! I did reach out the the alerts team to ask if they could make it so the informed entities stay in a consistent order across requests, so we would be able to get rid of the sorting once they implement that. But it once we have the summary caching in place, it makes sense to reconsider if we can get away with using timestamps instead.

lib/mobile_app_backend_web/channels/alerts_channel.ex
KaylaBrady: Ah true, tracking by process at that level would I think pretty much be something like the API's [DiffServer](https://github.com/mbta/api/blob/master/apps/api_web/lib/api_web/event_stream/diff_server.ex).

I am tempted to say that it would be safe to always send the diff to everyone who is subscribed to the same fetch keys, but I am not sure enough of that... is it possible for a user to miss an update if it happens while they are in the middle of subscribing? If that is possible, this approach does at least ensure that user's will end up with the right state at the next update.

KaylaBrady: Though reading the doc for DiffServer

>  However, when multiple clients are subscribed to the same data,
  this results in a lot of duplicated diffing effort.
>  This new implementation does the diffing once, and sends the diffs to each client.

Perhaps that theoretical race condition isn't an issue? Or at least, it would be worth load-testing this to make sure diffing at the consumer level doesn't cause any performance issues under rush hour conditions

EmmaSimon: Hmm, yeah I'm not certain, I would hope that the pubsub subscribe would properly queue up any changes even if they happened immediately after the connection, but I don't know if we have a good way to test that. I think doing some load testing after we have all the pieces in place is a good idea, then we can reevaluate if there are any issues.

PR #571, opened 2026-06-30
feat: calculate summaries as their own PubSub

boringcactus:
### Summary

_Ticket:_ [Alert Summary Consolidation | Cache Generated Summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215452720852110)

This architecture seems like the most straightforward way to make sure the channels always get up to date summaries.

### Testing

Copied over unit tests from the existing `Alerts.PubSub` and updated as needed.

### DIFF
--- a/lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
+++ b/lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
@@ -96,7 +96,7 @@ defmodule MobileAppBackend.Alerts.WithSummaryPubSub do
     upstream = Application.get_env(:mobile_app_backend, Alerts.PubSub, Alerts.PubSub)
 
     upstream.subscribe(legacy_compatibility: false)
-    broadcast_timer(50)
+    broadcast_initial_timer()
 
     create_table_fn =
       Keyword.get(opts, :create_table_fn, fn ->
--- a/lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
+++ b/lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
@@ -58,7 +58,7 @@ defmodule MobileAppBackend.Alerts.WithSummaryPubSub do
       %{
         alerts_with_summaries:
           data
-          |> Map.get({locale, :card}, %{})
+          |> Map.get(locale, %{})
           |> filter_upcoming_single_tracking_alerts()
       }
     end
@@ -143,7 +143,7 @@ defmodule MobileAppBackend.Alerts.WithSummaryPubSub do
   end
 
   @typep alerts_with_summaries :: %{Alert.id() => AlertWithSummaries.t()}
-  @typep summary_key :: {locale :: String.t(), :notification | :card}
+  @typep summary_key :: locale :: String.t()
   @typep all_summaries :: %{summary_key() => alerts_with_summaries()}
 
   defp recalculate(ets_table, all_alerts \\ nil) do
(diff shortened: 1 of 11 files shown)


lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
```diff
@@ -0,0 +1,174 @@
...
+      !(alert.cause == :single_tracking && !Alert.active?(alert))
+    end)
+  end
+
+  @impl GenServer
+  def init(opts \\ []) do
+    upstream = Application.get_env(:mobile_app_backend, Alerts.PubSub, Alerts.PubSub)
+
+    upstream.subscribe(legacy_compatibility: false)
+    broadcast_timer(50)
```
KaylaBrady: suggestion(non-blocking): I see this call to `broadcast_timer` already done in each of the other pubsubs, but now that we are adding a 4th use of it, I think it would be at least naming this constant, or otherwise pulling this out into a common definition.

lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
```diff
@@ -0,0 +1,174 @@
...
+        {:new_alerts, %{alerts: all_alerts}},
+        %{last_dispatched_table_name: last_dispatched} = state
+      ) do
+    all_summaries = recalculate(last_dispatched, Map.values(all_alerts))
+    perform_broadcast(last_dispatched, all_summaries)
+    {:noreply, state}
+  end
+
+  defp perform_broadcast(last_dispatched, all_summaries) do
+    Registry.dispatch(MobileAppBackend.Alerts.Registry, @fetch_registry_key, fn entries ->
```
KaylaBrady, line 129: thought(non-blocking): Looks like much of this is shared across the 4 pubsubs too - might be worth making a follow-up to refactor more of this out into `MobileAppBackend.PubSub`

lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
```diff
@@ -0,0 +1,174 @@
...
+    :ets.insert(ets_table, {:all_summaries, all_summaries})
+    all_summaries
+  end
+
+  @spec build_all_summaries([Alert.t()]) :: all_summaries()
+  defp build_all_summaries(alerts) do
+    alerts_by_id = Map.new(alerts, &{&1.id, &1})
+
+    for locale <- Application.get_env(:mobile_app_backend, :locale_codes),
+        context <- [:notification, :card],
```
KaylaBrady: question: why do we need to include the notification context here too?

lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
boringcactus: Makes sense.

lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
boringcactus: When this was a more general cache, it might have been reasonable to eventually read from it in the notifications code, too, but now that it’s more specific to the alert channel, we aren’t likely to ever expect this to contain the notifications versions of summaries. Good catch.

lib/mobile_app_backend/alerts/with_summary_pub_sub.ex
boringcactus: [Made](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1216241757035727?focus=true).

KaylaBrady (approved):

PR #574, opened 2026-07-01
fix: Make upcoming stop closures use AffectedStops

EmmaSimon:
### Summary

_Ticket:_ [Fix template for downstream stop closure](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1215459184849991?focus=true)

I spent too much time rearranging the template strings to not include the "at" and only include the "at" in the location fragment, but then I realized that it was only happening for upcoming alerts because instead of correctly using `AffectedStops` it was falling back to the standard `SingleStop`, `SuccessiveStops`, etc logic. The check for closure alerts was only returning true for alerts that were active now, so I made it also return true for upcoming alerts. We can't remove the timeframe entirely because all clear summaries still need to use the standard location.

While I was in here, I also added support for ferry closure summaries.

### Testing

Added tests for broken logic, also checked running locally that the frontend was receiving the correct summaries

### DIFF
--- a/lib/mobile_app_backend/alerts/alert_summary.ex
+++ b/lib/mobile_app_backend/alerts/alert_summary.ex
@@ -190,9 +190,9 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
     end
   end
 
-  defp alert_location_closure(alert, affected_stops) do
-    alert.effect in [:station_closure, :stop_closure] and
-      affected_stops != [] and Alert.active?(alert)
+  defp alert_location_is_closure?(alert, affected_stops) do
+    alert.effect in [:dock_closure, :station_closure, :stop_closure] and
+      affected_stops != [] and (Alert.active?(alert) or Alert.active_soon?(alert))
   end
 
   @spec alert_location(Alert.t(), Stop.id(), 0 | 1, [RoutePattern.t()], GlobalDataCache.data()) ::
@@ -224,7 +224,7 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
       downstream = Enum.all?(affected_stops, &(&1.id != stop_id))
 
       cond do
-        alert_location_closure(alert, affected_stops) ->
+        alert_location_is_closure?(alert, affected_stops) ->
           %Location.AffectedStops{
             stops: Enum.map(affected_stops, fn stop -> stop.name end)
           }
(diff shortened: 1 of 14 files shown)


lib/mobile_app_backend/alerts/alert_summary.ex
```diff
@@ -190,9 +190,9 @@ defmodule MobileAppBackend.Alerts.AlertSummary do
     end
   end
 
-  defp alert_location_closure(alert, affected_stops) do
-    alert.effect in [:station_closure, :stop_closure] and
-      affected_stops != [] and Alert.active?(alert)
+  defp alert_location_is_closure?(alert, affected_stops) do
+    alert.effect in [:dock_closure, :station_closure, :stop_closure] and
+      affected_stops != [] and (Alert.active?(alert) or Alert.active_soon?(alert))
```
KaylaBrady, line 195: question(non-blocking): This was a good catch to make the location consistent from active and upcoming, but I'm wondering why this needs to check time frame at all?

lib/mobile_app_backend/alerts/formatted_alert.ex
```diff
@@ -455,6 +455,7 @@ defmodule MobileAppBackend.Alerts.FormattedAlert do
 
   defp summary_affected_mode(effect) do
     case effect do
+      :dock_closure -> gettext("Ferries")
```
KaylaBrady, line 458: 🙌 TY for noticing this!

KaylaBrady (approved):

lib/mobile_app_backend/alerts/alert_summary.ex
EmmaSimon: That was my first impulse too, but we can't remove the timeframe entirely because all clear summaries still need to use the standard location.

KaylaBrady: Ah yeah I didn't totally get that on first read but now see what you mean. I think maybe part of the confusing thing then comes down to the name of the function. If an alert is a dock closure, couldn't we always represent it with `AffectedStops`, but then just localize `AffectedStops` differently for an `AllClear` alert? That might not be as straightforward as I'd imagine though.

EmmaSimon: Yeah that would be possible, we don't currently have a way to know in `summary_location` if it's for an all clear or not, but we could pass that in. Alternately, we could make it so that the preposition is included in every type of location string like I did in my first fix, then the all clear or the closure would both be able to use the same "at multiple stops" or "at X, Y, and Z" location strings, and the "at" wouldn't live in "will not stop at".

PR #582, opened 2026-07-09
perf: Improve API query retry behavior

KaylaBrady:
### Summary

_Ticket:_ [Investigate 7/7 increased latency & timeouts](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216357676564476?focus=true)

We received a 74 logs with the message `retry: got exception, will retry in 1000ms, 3 attempts left`. It seems that the API was potentially having some issues around this timeframe, but we can at least shorten the retry loop to fail faster. 

There were also a number of `Handler :default switched from :sync to :drop mode`, indicating that logs were being dropped. This also turns off some noisy logs that we don't often need to reduce the log volume

### Testing

* Ran app locally, confirmed fetching data still works

### DIFF
--- a/lib/mbta_v3_api.ex
+++ b/lib/mbta_v3_api.ex
@@ -196,7 +196,7 @@ defmodule MBTAV3API do
     [
       base_url: Application.get_env(:mobile_app_backend, :base_url),
       api_key: Application.get_env(:mobile_app_backend, :api_key),
-      timeout: 10_000
+      timeout: 7_000
     ]
   end
 
--- a/lib/mbta_v3_api.ex
+++ b/lib/mbta_v3_api.ex
@@ -137,6 +137,8 @@ defmodule MBTAV3API do
           params: params,
           compressed: true,
           decode_body: false,
+          retry_delay: fn _ -> 300 end,
+          max_retries: 2,
           pool_timeout: timeout,
           receive_timeout: timeout
         )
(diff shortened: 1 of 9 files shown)


lib/mbta_v3_api.ex
```diff
@@ -137,6 +137,8 @@ defmodule MBTAV3API do
           params: params,
           compressed: true,
           decode_body: false,
+          retry_delay: fn _ -> 300 end,
+          max_retries: 2,
```
KaylaBrady, line 141: I thought about also configuring `request_timeout`, which [defaults to infinity ](https://finch.hexdocs.pm/Finch.html#request/3-options). We could set it to something safely large like 5 seconds?

KaylaBrady:
Something else I considered but didn't end up including in that PR was decreasing the max_concurrency when fetching schedules in parallel to something more like 10. 

I originally set it to 25 thinking that we could take advantage of our large HTTP connection pools, but I don't know that we ever actually check out more than 10 connections at once. We are doing more than just making HTTP requests in parallel (caching, post-processing data), so 25 could be too aggressive, and 10 is still above our avg number of stops.

But I could see waiting to make that change if we continue to see issues with this endpoint (we originally increased the max_concurrency as an attempt to fix timeout errors, so I don't have super high expectations that lowering it again will help outright)

lib/mbta_v3_api.ex
EmmaSimon: I think we should yeah, if a request takes that long, something is probably wrong.

EmmaSimon (approved):

PR #597, opened 2026-07-15
Add logging when trip subscription fails

thedecaftony:
### Summary

_Ticket:_ [Investigate Adam's report of GL-C weirdness](https://app.asana.com/1/15492006741476/project/1215456342910576/task/1215638932339487?focus=true)

Add more details logging to be able to spot why REFUSED join is happening

### Testing

Deployed to dev orange and check that the logs existed

[Splunk logs](https://mbta.splunkcloud.com/en-US/app/search/show_source?sid=1784128001.49472&offset=0&latest_time=)
```
94fdb785c6f1 13:17:22.921 [warning] Elixir.MobileAppBackend.Predictions.StreamSubscriber.Impl failed to fetch trip from repository for trip=ADDED-1748030133 not found.
```

### DIFF
--- a/lib/mobile_app_backend/predictions/stream_subscriber.ex
+++ b/lib/mobile_app_backend/predictions/stream_subscriber.ex
@@ -66,7 +66,8 @@ defmodule MobileAppBackend.Predictions.StreamSubscriber.Impl do
   @impl true
   def subscribe_for_trip(trip_id) do
     with {:ok, %{data: [trip]}} <- MBTAV3API.Repository.trips(filter: [id: trip_id]),
-         route_id <- trip.route_id,
+         {:ok, route_id} <-
+           if(trip.route_id, do: {:ok, trip.route_id}, else: {:error, :missing_route_id}),
          {:ok, _data} <-
            StaticInstance.ensure_stream_started("predictions:route:to_store:#{route_id}",
              include_current_data: false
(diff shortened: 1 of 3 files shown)


lib/mobile_app_backend/predictions/stream_subscriber.ex
```diff
@@ -66,7 +66,8 @@ defmodule MobileAppBackend.Predictions.StreamSubscriber.Impl do
   @impl true
   def subscribe_for_trip(trip_id) do
     with {:ok, %{data: [trip]}} <- MBTAV3API.Repository.trips(filter: [id: trip_id]),
-         route_id <- trip.route_id,
+         {:ok, route_id} <-
+           if(trip.route_id, do: {:ok, trip.route_id}, else: {:error, :missing_route_id}),
```
KaylaBrady, line 70: question: When would route_id be null, have you seen this in practice? Our data types on the frontend & backend both have route_id as a required field for trip, so I'd expect that we always have one (or would break when parsing the data earlier if not)

lib/mobile_app_backend/predictions/stream_subscriber.ex
```diff
@@ -75,8 +76,29 @@ defmodule MobileAppBackend.Predictions.StreamSubscriber.Impl do
            StaticInstance.ensure_stream_started("vehicles:to_store", include_current_data: false) do
       :ok
     else
+      {:ok, %{data: []}} ->
+        Logger.warning(
+          "#{__MODULE__} failed to fetch trip from repository for trip=#{trip_id} not found."
```
KaylaBrady: suggestion(non-blocking): maybe worth a little helper function to contain the common log construction

lib/mobile_app_backend_web/channels/predictions_for_trip_channel.ex
```diff
@@ -21,11 +21,14 @@ defmodule MobileAppBackendWeb.PredictionsForTripChannel do
 
     case :timer.tc(fn -> pubsub_module.subscribe_for_trip(trip_id) end) do
       {time_micros, :error} ->
-        Logger.warning("#{__MODULE__} failed join duration=#{time_micros / 1000}")
+        Logger.warning(
+          "#{__MODULE__} failed join trip_id=#{trip_id} duration=#{time_micros / 1000}"
```
KaylaBrady, line 25: 👍 appreciate you adding trip_id in a searchable way here!

lib/mobile_app_backend/predictions/stream_subscriber.ex
thedecaftony: I have not seen it happening, but I wanted to see if I could make the logging more explicit in case we saw it happening

lib/mobile_app_backend/predictions/stream_subscriber.ex
thedecaftony: Oh you're right, that would reduce this logic by a lot, let me do it now

KaylaBrady (approved):

PR #608, opened 2026-07-16
fix: Fix elixir/erlang renovate issues

EmmaSimon:
### Summary

_Ticket:_ No ticket, follow up to #595 

This updates a few different renovate issues:
* If no erlang version is specified, the renovate mix manager will default to erlang 26, which is not compatible with elixir 1.20
* For regex managers, `fileMatch` is deprecated, but I was also using `managerFilePatterns` wrong.
* Added a new manager to update the elixir version in the mix project, as far as I can tell, the only thing this will do is prevent `mix compile` from succeeding if you have the wrong elixir version installed

### Testing

Won't know for sure if this works until this is merged and runs on mend

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -5,7 +5,8 @@ defmodule MobileAppBackend.MixProject do
     [
       app: :mobile_app_backend,
       version: "0.1.0",
-      elixir: "~> 1.14",
+      # renovate: datasource=github-releases depName=elixir-lang/elixir
+      elixir: "1.20.2",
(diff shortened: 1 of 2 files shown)


KaylaBrady (approved):

KaylaBrady:
TY for following up on this! I think it would be worth sharing your config customizations here with the wider eng department once this in place!

EmmaSimon:
> TY for following up on this! I think it would be worth sharing your config customizations here with the wider eng department once this in place!

We should probably wait for an update to make sure they actually work first 😅

PR #611, opened 2026-07-17
perf?: Increase response cache size

KaylaBrady:
### Summary

_Ticket:_ [Improve caching](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216655729859154?focus=true)

We observed a high miss & eviction rate in the ResponseCache, which caches responses directly from the V3API and re-uses them to fetch with the `if-modified-since` header. Increasing the cache size should help our hit rate, which we hypothesize will help with some ongoing latency issues. 
image

This additionally adds the gc_interval for the cache. [Per the docs](https://nebulex-local.hexdocs.pm/Nebulex.Adapters.Local.html#module-configuration-options), "Always provide the :gc_interval option so the garbage collector can work appropriately out of the box."

### Testing

* Will continue to monitor the hit rate dashboard.
* Added unit tests for new telemetry events & ran locally to confirm it logged as expected

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -33,12 +33,14 @@ config :mobile_app_backend, vehicles_broadcast_interval_ms: 500
 config :mobile_app_backend, MBTAV3API.ResponseCache,
   gc_interval: :timer.hours(12),
   allocated_memory: 2_000_000_000,
-  stats: true
+  stats: true,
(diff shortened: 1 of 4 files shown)


EmmaSimon (approved):
🎉

PR #614, opened 2026-07-20
Revert "chore(deps): update dependency elixir to v1.20.2 (#595)"

KaylaBrady:
This reverts commit bfd7bc6c55a5e4878ec24f601a34f6f3935c1856.

### Summary

_Ticket:_ No ticket 

Temporarily reverts https://github.com/mbta/mobile_app_backend/pull/595. On deploying 2026-07-20-1, there were issues with the GlobalDataCache not returning 304s at the rate we would expect and . Since this upgrade touched the GlobalDataCache (though in a seemingly no-op way). 

We also saw tasks fail with the error `ea42114edec0 sys/unix/sys_signal_stack.c:103:sys_sigaltstack(): Internal error: Failed to set alternate signal stack`. 

It looks like there is a fix coming in a future version for this issue: 
https://github.com/erlang/otp/issues/11248

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -5,8 +5,7 @@ defmodule MobileAppBackend.MixProject do
     [
       app: :mobile_app_backend,
       version: "0.1.0",
-      # renovate: datasource=github-releases depName=elixir-lang/elixir
-      elixir: "1.20.2",
+      elixir: "~> 1.14",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
       aliases: aliases(),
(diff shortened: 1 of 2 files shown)


boringcactus (approved):

boringcactus:
Seems like one or two pieces of https://github.com/mbta/mobile_app_backend/pull/608 may need rolling back, too.

boringcactus:
Looks like a force push replaced reverting #595 with reverting #608 instead of reverting both.

k-patrik:
Hi @KaylaBrady! A friendly suggestion to pin the Alpine version to `3.23` as an alternative. That mitigates the sigaltstack rejections without the need for downgrading BEAM. Compete set of details in https://github.com/erlang/otp/pull/11376

PR #630, opened 2026-07-24
chore: ditch abandoned GCP client libraries

boringcactus:
### Summary

_Ticket:_ none
[Slack thread](https://mbta.slack.com/archives/C062QNAJZ2M/p1784920942210449)

Unfortunately, Google has for some reason decided to [deprecate the Elixir client libraries for Google Cloud Platform](https://github.com/googleapis/elixir-google-api#%EF%B8%8F-deprecation-notice-%EF%B8%8F) and archive the repository, so this PR rebuilds the pieces we actually use.

### Testing

Updated tests to ensure that behavior remains the same (except for a few things in the Firebase send call that were documented as snake_case but implemented by the old Elixir client as camelCase, where I’ve chosen to follow the documentation instead, since snake_case is idiomatic Elixir). Checked on dev-orange that notification delivery isn’t completely broken.


lib/util/gcp/fcm.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  @doc "https://firebase.google.com/docs/reference/fcm/rest/v1/projects.messages/send"
+  @spec send(String.t(), String.t(), %{
+          optional(:validate_only) => boolean() | nil,
+          message: Message.t()
+        }) ::
+          {:ok, Req.Response.t()} | {:error, Exception.t()}
+  def send(token, parent, body) do
+    Util.GCP.base_req()
+    |> Req.post(
```
KaylaBrady, line 82: question(non-blocking): Do we need any custom configuration for req pooling at the volume we're sending? I think since the oban queue concurrency default is set to 10 this wouldn't actually be an issue, but maybe something to look at when you test in dev orange

lib/util/gcp.ex
```diff
@@ -0,0 +1,10 @@
+defmodule Util.GCP do
+  @moduledoc """
+  Wrappers for Google Cloud Platform APIs to fill in for the inexplicably-deprecated
+  [official Google client libraries](https://github.com/googleapis/elixir-google-api).
+  """
+
+  def base_req do
+    Req.new(Application.get_env(:mobile_app_backend, __MODULE__, []))
```
KaylaBrady, line 8: 👍 this is cool, haven't seen this before

KaylaBrady (approved):
Appreciate how you organized this! Just a few minor comments, but if all looks good on dev-orange then LGTM!

lib/util/gcp/fcm.ex
boringcactus: Since the default pool is separate per host, and the Oban concurrency is not particularly high, we should be fine. We want to avoid getting rate limited by the FCM API, anyway.

boringcactus:
Checked on dev-orange and it’s working!

PR #636, opened 2026-07-27
chore: update test data

boringcactus:
### Summary

_Ticket:_ none

It’s been a while since we’ve updated the test data, and I’m looking at using the test data snapshots for testing the SummaryEntityBuilder. Splitting this out into its own PR seems like it’ll make following changes easier to review.

### Testing

Checked that tests still pass.


KaylaBrady (approved):
👍 I'm still generally in favor of minimizing the places where we use this low-level mocked data, but since the relationships between the data are so important for summaries as you've pointed out, this does seem like a good use case for it.

PR #637, opened 2026-07-27
refactor: use real data for SummaryEntityBuilderTest

boringcactus:
### Summary

_Ticket:_ [Alert Summary Consolidation | Backend generates all necessary summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216565516165771)

Knowing we’re correctly handling downstream alert summary generation will be a little tricky; using real data about the system instead of fake data makes it easier for us to think about downstreamhood and makes sure the tests aren’t unrealistically missing things like downstream stops that real data will definitely have. Changing the tests before changing the behavior ensures that we’re actually testing the cases we want to be testing.

### Testing

Checked that tests still pass.


EmmaSimon (approved):
Good call!

PR #640, opened 2026-07-28
feat: build downstream alert summaries

boringcactus:
### Summary

_Ticket:_ [Alert Summary Consolidation | Backend generates all necessary summaries](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216565516165771)

This wound up being more complicated than I thought, and I didn’t exactly think it would be easy.

I do not think there are other cases we’re missing.

### Testing

Updated unit tests to expect new behavior.

### DIFF
--- a/lib/mobile_app_backend/alerts/summary_entity_builder.ex
+++ b/lib/mobile_app_backend/alerts/summary_entity_builder.ex
@@ -1,4 +1,5 @@
 defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
+  require Logger
   alias MBTAV3API.Alert
   alias MBTAV3API.Repository
   alias MBTAV3API.Route
@@ -10,17 +11,6 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
   alias MobileAppBackend.Alerts.FormattedAlert
   alias MobileAppBackend.Alerts.SummaryEntity
   alias MobileAppBackend.GlobalDataCache
-
-  defmodule BroadCombination do
-    @moduledoc "a route, direction, and optional trip that an InformedEntity applies to"
-    @type t :: %__MODULE__{
-            route: Route.id(),
-            direction: 0 | 1,
-            trip: Trip.id() | nil
-          }
-    @enforce_keys [:route, :direction, :trip]
-    defstruct [:route, :direction, :trip]
-  end
 
   defmodule Combination do
     @moduledoc "specific parameters for an individual summary"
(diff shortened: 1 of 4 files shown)


lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -93,130 +117,191 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
...
+    alert.informed_entity
+    |> Enum.flat_map(&combination_from_entity(&1, {all_child_stops, trips, global}))
+    |> Enum.uniq()
+    |> Enum.flat_map(&expand_combination(&1, global))
+  end
 
-    # Filter out any combinations that are already covered by another wildcard in the list
-    Enum.reject(combinations, fn combination ->
-      redundant_combination?(combination, combinations)
+  @spec combination_from_entity(
```
KaylaBrady: suggestion(non-blocking): I think it may help for readability to do some named guards here instead of both pattern matching & when clauses in each. This doesn't exactly match [complex extraction in clauses](https://elixir.hexdocs.pm/code-anti-patterns.html#complex-extractions-in-clauses), but I think it is getting complex enough to warrant some simplifying helpers!

```
def combination_from_entity(entity, extra) when is_only_route_type(entity) do
   ...
end

def combination_from_entity(entity, extra) when is_trip(entity) do
   ...
end
        
```

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -93,130 +117,191 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
...
+         %BroadCombination{route: route, direction: direction, trip: trip},
+         global
+       ) do
+    patterns = RoutePattern.get_relevant_patterns(route, nil, direction, global)
+
+    # we don’t actually need to look at each stop separately if
+    # 1. this isn’t a trip-specific alert
+    # 2. this isn’t on the Green Line
+    # 3. there’s only one pattern in this direction, so all the stops are on it
+    should_expand_stops? =
```
KaylaBrady: question: Could we get rid of the BroadCombination data type entirely if we moved this check out of expand_combination and before combination_from_entity?

Thinking something like 
```
def combinations_from_entity do
    expanded_entities = if (should_expand_stops) do
      map single informed_entity into a list of {entity, patterns}
   else
     [{entity, patterns}]
   end
  
  Enum.flat_map(expanded_entities, fn {entity, patterns} -> combination_from_entity(entity,
                                                                    {all_child_stops, trips_global, patterns} end)
end
```

lib/mobile_app_backend/alerts/summary_entity_builder.ex
```diff
@@ -337,16 +422,28 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilder do
...
-      case grouped do
-        [a, b] when a.summary == b.summary ->
-          [%{a | direction_id: nil}]
-
-        _ ->
-          grouped
-      end
-    end)
+    # sorted from least desirable to split by to most desirable to split by
+    all_keys = [:stop_id, :trip_id, :direction_id, :route_id]
```
KaylaBrady: question: I remember you describing this at dev check-in, but I'm not totally following why we need this or what makes :stop_id less desirable than :route_id

test/mobile_app_backend/alerts/summary_entity_builder_test.exs
```diff
@@ -91,58 +119,339 @@ defmodule MobileAppBackend.Alerts.SummaryEntityBuilderTest do
...
+                 direction_id: nil,
+                 trip_id: nil,
+                 summary: "12:06 PM train from Andrew is suspended today due to maintenance"
+               },
+               %SummaryEntity{
+                 route_id: nil,
+                 stop_id: "place-asmnl",
+                 direction_id: nil,
+                 trip_id: nil,
+                 summary: "Service suspended on Red Line"
```
KaylaBrady: question: why doesn't this one have a time specified? I know the original test was like this too, but I'm missing something about why we'd have a mix of generic suspension and trip specific suspension messages, both with `trip_id` nil (`trip_id` nil feels weirder on the other one since that at least mentions a specific train)

lib/mobile_app_backend/alerts/summary_entity_builder.ex
KaylaBrady: suggestion: With the new complexity, I think it would be worth making this doc a bit more detailed, including the new behavior for downstream and what we discussed at dev check-in about how the nil / wildcards are still within the context of the alert, and we rely on other logic to determine where the alert should be shown

lib/mobile_app_backend/alerts/summary_entity_builder.ex
boringcactus: Good call. There are more cases than there should be, too, because I added extra cases to the original `case` to handle both specified and unspecified directions in the route and trip cases, and then I turned those into separate function clauses despite that no longer making any sense.

lib/mobile_app_backend/alerts/summary_entity_builder.ex
boringcactus: Expanding an `InformedEntity` to add stop IDs to it that it doesn’t actually apply to feels weird to me, but expanding wildcards in `route` and `direction` in an `expand_entity` feels better, and then `combinations_from_entity` only needs to expand the stops. At that point, I think guards make `expand_entity` less legible.

lib/mobile_app_backend/alerts/summary_entity_builder.ex
boringcactus: Strictly speaking, we could skip this entire function and just rely on compression to reduce the cost of redundant summaries, but I think it’s worth having at least some attempt to deduplicate our summaries. If every summary of this alert is the same, we only want to have one copy of it, and if there are only a few different cases, we want as few copies as possible. If we can either have an entity per stop with wildcards in routes or an entity per route with wildcards in stops, we would prefer to collapse the stops and keep the routes as separate, because there will be fewer distinct routes than distinct stops. I think I can clear up the comments here somewhat.

lib/mobile_app_backend/alerts/summary_entity_builder.ex
boringcactus: Added some detail.

test/mobile_app_backend/alerts/summary_entity_builder_test.exs
boringcactus: The alert only applies to one trip, so the trip ID is not needed to distinguish between summary entities, and therefore `dedup_summaries/1` nils it out. The trip is on the Braintree branch, so from the Ashmont branch, there is no time to show, so the trip-specific alert case does not fire at all; the trip will not be shown on the Ashmont branch, so this summary will never be displayed.

It might make sense to only generate summaries for trip-specific alerts at the stops that those trips are scheduled to visit. The only problem case I can think of is if a trip with a trip-specific alert is predicted to make an unscheduled stop, but that seems extremely unlikely to happen, and if it does, the frontend’s fallback case for a missing summary (which is to just show the alert effect in the card) seems better than this scenario.

lib/mobile_app_backend/alerts/summary_entity_builder.ex
KaylaBrady: That helps!

lib/mobile_app_backend/alerts/summary_entity_builder.ex
KaylaBrady: Yeah I like that!

KaylaBrady (approved):

PR #645, opened 2026-07-30
chore: pin alpine to 3.23

KaylaBrady:
### Summary

_Ticket:_ None

Fix the `Internal error: Failed to set alternate signal stack` error that https://github.com/mbta/mobile_app_backend/pull/619 did not. TY @k-patrik for the suggestion!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # --- Set up Elixir build ---
 ARG ELIXIR_VERSION=1.20.2
 ARG ERLANG_VERSION=28.5.0.3
-ARG ALPINE_VERSION=3.24.1
+ARG ALPINE_VERSION=3.23.4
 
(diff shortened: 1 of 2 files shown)


EmmaSimon:
Looks like the only images that exist in docker hub with that elixir and otp combo have Alpine 3.23.5

Dockerfile
```diff
@@ -1,7 +1,7 @@
 # --- Set up Elixir build ---
 ARG ELIXIR_VERSION=1.20.2
-ARG ERLANG_VERSION=28.5.0.3
-ARG ALPINE_VERSION=3.24.1
+ARG ERLANG_VERSION=29.0.4
```
boringcactus, line 3: This should probably be kept in sync with `.tool-versions`.

boringcactus (approved):

PR #655, opened 2026-07-31
feat: support FCM installation ID alongside old token

boringcactus:
### Summary

_Ticket:_ [Move from Firebase tokens to installation IDs](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216241757035733)

This seems like the most reasonable way to handle both kinds of target. Pairs with https://github.com/mbta/mobile_app/pull/1911.

### Testing

Added and updated unit tests to ensure functionality with both kinds of target. End-to-end delivery test still needs to be done.


lib/util/fcm_target.ex
```diff
@@ -0,0 +1,37 @@
+defmodule Util.FCMTarget do
```
KaylaBrady, line 1: 👍 I like this approach

lib/mobile_app_backend/notifications/deliverer.ex
```diff
@@ -43,6 +43,15 @@ defmodule MobileAppBackend.Notifications.Deliverer do
         _ -> alert_id
       end
 
+    {installation_id, token} =
+      case user do
+        %User{fcm_installation_id: installation_id} when not is_nil(installation_id) ->
+          {installation_id, nil}
+
+        %User{fcm_installation_id: nil, fcm_token: token} ->
```
KaylaBrady, line 51: question(non-blocking): Do we ~always have to support this path since users may not upgrade their app? Or is there a point where the old token just won't work anymore, and users might stop getting notifications without knowing why?

KaylaBrady (approved):

lib/mobile_app_backend/notifications/deliverer.ex
boringcactus: One day, Google will probably shut off the old notification delivery path, and our users on old versions of the app will stop getting notifications. That will probably not happen anytime soon (even Google probably won’t do it in less than a year), and we should have plenty of notice ahead of time, so we could probably send out a targeted “please upgrade your app, you are on an old version and notifications will break soon” notification if we needed to.

boringcactus:
Things are [busted](https://mbta.slack.com/archives/C05QMG9GS9M/p1786046272829149); this is likely to be a frontend issue rather than a backend issue, but it’s hard to know, and it’s bad to guess.

PR #657, opened 2026-07-31
feat: Introduce Keycloak login for admin dashboards

EmmaSimon:
### Summary

_Ticket:_ [Enable LiveDashboard](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216655729859151?focus=true)

Corresponding https://github.com/mbta/devops/pull/4703 and https://github.com/mbta/terraform_modules/pull/943 PRs

This adds a new lightweight `/dev` endpoint to the backend which links out to the Phoenix LiveDashboard and the Oban dashboard. Since we want to view these in prod, access needed to be limited, so Keycloak auth was added to anything in the `/dev` path. 

Most of the auth logic was pulled from https://github.com/mbta/alerts_ui, with some tweaks because our use case is much simpler (like not needing to support logging out).

I also fixed an issue we've had for a while where tailwindcss was not properly loading styles.

Once this is approved we'll need a devops PR to update the image in staging.

### Testing

Deployed to dev orange and verified that `/dev`, `/dev/dashboard`, and `/dev/oban` are only accessible after logging in.


lib/mobile_app_backend/keycloak_user.ex
```diff
@@ -0,0 +1,64 @@
...
+  @type role :: :developer
+  @type t :: %__MODULE__{
+          keycloak_id: String.t(),
+          email: String.t(),
+          first_name: String.t(),
+          last_name: String.t(),
+          roles: [role()]
+        }
+
+  schema "users" do
```
KaylaBrady, line 17: question: Is this and the corresponding changeset used at all?

lib/mobile_app_backend_web/router.ex
```diff
@@ -1,20 +1,37 @@
...
 
   pipeline :browser do
     plug :accepts, ["html"]
+    plug :redirect_prod_http
     plug :fetch_session
     plug :fetch_live_flash
     plug :put_root_layout, html: {MobileAppBackendWeb.Layouts, :root}
     plug :protect_from_forgery
     plug :put_secure_browser_headers
+    plug :fetch_current_user
```
KaylaBrady: question: Why include this at the top level and not within the dev path?

lib/mobile_app_backend_web/router.ex
```diff
@@ -57,22 +74,30 @@ defmodule MobileAppBackendWeb.Router do
...
-
-      live_dashboard "/dashboard", metrics: MobileAppBackendWeb.Telemetry
-      oban_dashboard("/oban")
-    end
+  scope "/dev", MobileAppBackendWeb do
+    pipe_through [:browser, :require_authenticated_user, :require_developer]
+
+    get "/", DevController, :home
+    live_dashboard "/dashboard", metrics: MobileAppBackendWeb.Telemetry
+    oban_dashboard("/oban")
```
KaylaBrady, line 81: Oh sweet, I didn't realize we get an oban dashboard too!

lib/mobile_app_backend_web/user_auth.ex
```diff
@@ -0,0 +1,269 @@
...
+    last_active = get_session(conn, :last_active, nil)
+
+    conn
+    |> assign(:current_user, user)
+    |> assign(:last_auth_time, last_auth_time)
+    |> assign(:last_active, last_active)
+  end
+
+  @doc """
+  Handles mounting and authenticating the current_user in LiveViews.
```
KaylaBrady, line 54: question(non-blocking): do we need this?

KaylaBrady (commented):
Awesome! Just a few tiny Qs

lib/mobile_app_backend_web/user_auth.ex
EmmaSimon: I started pulling it out, then realized that LiveDashboard definitely uses it and Oban probably does, and I didn't want to cause any problems with those.

lib/mobile_app_backend/keycloak_user.ex
EmmaSimon: Oh, no it shouldn't be, good catch.

lib/mobile_app_backend_web/router.ex
EmmaSimon: I was trying to streamline things, but you're right, it should only be necessary on dev

KaylaBrady (approved):
🎉

PR #668, opened 2026-08-05
feat(WithSummaryPubSub): Re-enable, and recalculate only summaries that may have changed

KaylaBrady:
### Summary

_Ticket:_ [Alert Summary Consolidation | Load testing](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216122497530220?focus=true)

While monitoring LiveDashboard in staging, I noticed WithSummaryPubSub had a MsgQ that kept creeping up, ~1 per minute. After adding some temporary loging, I found it takes ~50 seconds to recalculate all summaries. There is a timed recalculation every 60 seconds, plus whenever the alerts data updates, which seems to also happen every ~60 seconds (I think because "until further notice" alerts get extended end periods).

After this change, we only recalculate summaries for alerts that plausibly may have a summary change

### Testing
* Added unit tests
* Deployed to staging and confirmed not every summary is recalculated
image

### Notes
* This wasn't caught when load testing in dev-orange because the alerts in dev-orange are different than prod. For alerts-related load testing, we may want to either make more realistic alerts or temporarily switch over to using a V3 API environment that includes the prod alerts feed.

### DIFF
--- a/lib/mobile_app_backend/alerts/alert_with_summaries.ex
+++ b/lib/mobile_app_backend/alerts/alert_with_summaries.ex
@@ -1,7 +1,5 @@
 defmodule MobileAppBackend.Alerts.AlertWithSummaries do
   alias MBTAV3API.Alert
-  alias MBTAV3API.Alert.ActivePeriod
-  alias MBTAV3API.Alert.InformedEntity
   alias MobileAppBackend.Alerts.SummaryEntity
 
   @type t :: %__MODULE__{
--- a/lib/mobile_app_backend/alerts/alert_with_summaries.ex
+++ b/lib/mobile_app_backend/alerts/alert_with_summaries.ex
@@ -41,4 +41,15 @@ defmodule MobileAppBackend.Alerts.AlertWithSummaries do
       Util.DateTime.datetime_to_gtfs(old_alert_with_summaries.summaries_updated_at) !=
         Util.DateTime.datetime_to_gtfs(now)
   end
+
+  @spec flatten_alert_fields(t()) :: map()
+  @doc """
+  Flattens the alert fields to the top level for convenience when treating an `AlertWithSummaries` as an `Alert`
+  """
+  def flatten_alert_fields(alert_with_summaries) do
+    alert_with_summaries
+    |> Map.from_struct()
+    |> Map.drop([:alert])
+    |> Map.merge(Map.from_struct(alert_with_summaries.alert))
+  end
 end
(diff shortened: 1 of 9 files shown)


lib/mobile_app_backend/alerts/alert_with_summaries.ex
```diff
@@ -38,10 +39,37 @@ defmodule MobileAppBackend.Alerts.AlertWithSummaries do
...
+  - The alert has changed (e.g. description, header, effect, etc.)
+  - The alert has changed from active to inactive or vice versa
+  - It is a different day than the last time summaries were calculated (either calendar date or service date)
+  """
+  def should_recalculate_summaries?(
+        old_alert_with_summaries,
+        new_alert,
+        now \\ DateTime.now!("America/New_York")
+      ) do
+    old_alert = struct(Alert, Map.from_struct(old_alert_with_summaries))
```
KaylaBrady: This I don't like... really wish there was a nicer way to represent inheritance

EmmaSimon: Oh, yeah that's kind of unpleasant. Could we make all of the utility functions in `Alert` take both an `Alert` or an `AlertWithSummaries`, or is that too weird? Or make a separate 3rd module that contains the functions which accepts either? But I guess there would still need to be an equality check somewhere that still did this... So maybe this is the best approach.

EmmaSimon (approved):
Ok nice, this isn't bad at all!

lib/mobile_app_backend/alerts/alert_with_summaries.ex
KaylaBrady: @EmmaSimon while playing around with your suggestion for functions to take an `Alert` or `AlertWithSummaries`, I was tempted to try out composition and just give `AliertWithSummaries` a distinct `alert` field. I think it came out pretty tidy, and failing tests prompted simplification of the `alert_hashes` function now that we have a `summaries_updated_at` time. I'll hold of on merging until you have a chance to review this change!

EmmaSimon: Yeah I like this, but I'm not sure how annoying it will make parsing on the frontend. I think we'd need to write a custom deserializer to parse it into an `Alert` object. We could just go with this same structure on the frontend, but that's backwards incompatible and would require rewriting every place in the frontend that uses an alert, which I'd love to avoid... Or we could convert it into a more alert shaped structure on the backend before sending it.

Nice one with the `alert_hashes`, I was hoping we'd be able to get rid of that before we release this. 🙏

KaylaBrady: Oh true... I can add the flattening of the alert at the channel level since I'm betting that will end up cleaner than the custom deserializer

EmmaSimon (approved):

PR #684, opened 2026-08-13
fix(Notifications.Engine): FH2 trip-specific alert missing location & timeframe

KaylaBrady:
### Summary

_Ticket:_ [Trip-specific ferry alert missing information](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217168623666487?focus=true)

There were 2 issues involved here:
1. Users with Hull favorited would receive a misleading Standard notification like "Ferries will not stop at Georges Island starting 10:15 AM today" even though the trip specified in the alert doesn't serve Hull. This is misleading, because only 1 trip for the day was skipping Georges, not all trips after 10:15. They should have received no notification.
2. Users with Hull **and** Hingham favorited would receive a very confusing notification like ""Ferries will not stop at  " because the logic for combining summaries doesn't really work for combing a Standard notification (wrong, from Hull) with a trip-specific notification (right, from Hingham). 

This PR addresses the core of 1, which makes 2 not possible for this scenario. However, it is worth further consideration how we attempt to combine summaries, which I have made a [placeholder ticket
](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1217454774712131?focus=true) for.
### Testing

* Added unit tests
* ~TODO:~ test in dev-orange

### DIFF
--- a/lib/mobile_app_backend/notifications/engine.ex
+++ b/lib/mobile_app_backend/notifications/engine.ex
@@ -82,6 +82,8 @@ defmodule MobileAppBackend.Notifications.Engine do
           [subscription.stop_id]
       end
 
+    alerts = filter_trip_alerts_serving_stop(alerts, target_stop_with_children)
+
     applicable_alerts =
       applicable_alerts(alerts, subscription, route_ids, target_stop_with_children)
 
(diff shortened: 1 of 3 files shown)


lib/mobile_app_backend/notifications/engine.ex
```diff
@@ -82,6 +82,8 @@ defmodule MobileAppBackend.Notifications.Engine do
           [subscription.stop_id]
       end
 
+    alerts = filter_trip_alerts_serving_stop(alerts, target_stop_with_children)
```
KaylaBrady, line 85: With this and schedules both being trip specific, I may try a follow up refactor to separate the trip-specific path out more specifically.

boringcactus (approved):

PR #685, opened 2026-08-13
Chore: Add logs to track longest backend calls perfomance

thedecaftony:
### Summary

_Ticket:_ [re-assess backend stability ](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1216836286348078?focus=true)

### Testing

- Ran unit tests

### DIFF
--- a/lib/mbta_v3_api/stream/instance.ex
+++ b/lib/mbta_v3_api/stream/instance.ex
@@ -90,9 +90,9 @@ defmodule MBTAV3API.Stream.Instance do
       |> Enum.map_join(" ", fn {name, value} -> "#{name}=#{value}" end)
 
     if stage_healthy and consumer_healthy do
-      Logger.info("#{__MODULE__} #{health_state}")
+      Logger.debug("#{__MODULE__} #{health_state} consumer_pid=#{inspect(consumer_pid)}")
     else
-      Logger.warning("#{__MODULE__} #{health_state}")
+      Logger.warning("#{__MODULE__} #{health_state} consumer_pid=#{inspect(consumer_pid)}")
     end
   end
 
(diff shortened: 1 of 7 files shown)


lib/mobile_app_backend_web/controllers/schedule_controller.ex
```diff
@@ -12,7 +12,16 @@ defmodule MobileAppBackendWeb.ScheduleController do
     if stop_ids_concat == "" do
       json(conn, %{schedules: [], trips: %{}})
     else
-      fetch_schedules_and_combine_data(conn, stop_ids_concat, date_time_string, params)
+      {time_in_us, result} =
+        :timer.tc(fn ->
+          fetch_schedules_and_combine_data(conn, stop_ids_concat, date_time_string, params)
+        end)
+
+      Logger.info(
```
KaylaBrady: question: Is this redundant to the log below? if not, maybe worth adding something unique to distinguish it from the other "schedules request" log

lib/mbta_v3_api/stream/instance.ex
```diff
@@ -90,9 +90,9 @@ defmodule MBTAV3API.Stream.Instance do
       |> Enum.map_join(" ", fn {name, value} -> "#{name}=#{value}" end)
 
     if stage_healthy and consumer_healthy do
-      Logger.info("#{__MODULE__} #{health_state}")
```
KaylaBrady, line 93: suggestion (non-blocking): this is a function that [timed out in the past](https://app.asana.com/1/15492006741476/project/1205425564113216/task/1216662100918721?focus=true), so we have some signal that there might actually be performance issues here. It could be interesting to add duration to these existing logs & keep it at the `info` level to motivate whether to prioritize the linked ticket.

KaylaBrady (commented):
👍 a few minor comments!

lib/mobile_app_backend_web/controllers/schedule_controller.ex
thedecaftony: I missed that, it actually is, let me remove it

lib/mbta_v3_api/stream/instance.ex
thedecaftony: Makes sense Iw as trying to remove a log that had a log events

KaylaBrady (approved):

PR #686, opened 2026-08-13
fix: Get schedules for future trip alerts

EmmaSimon:
### Summary

_Ticket:_ [Trip-specific alert for trip on a different day is missing information](https://app.asana.com/1/15492006741476/project/1205732265579288/task/1217168623666489?focus=true)

Fetch schedules in the future for trip specific alerts. Requests will be made for today and/or tomorrow, depending on the alert's active period. If we update summaries in the future to support displaying them more than one day in advance, we'll need to handle this differently. These changes apply to both notifications and summary entity builds, and the trip schedule fetching code for both has been consolidated into a new `AlertUtil` module.

### Testing

Added a unit test to verify that requests are made for future days if the alert's trips are not found for the current day, updated summary builder tests to handle new behavior


lib/mobile_app_backend/alerts/alert_util.ex
```diff
@@ -0,0 +1,160 @@
...
+              acc_schedules ++ date_schedules,
+              Map.merge(acc_trips, date_trips)
+            }
+
+          response ->
+            Logger.error(
+              "failed to fetch schedules for trip_ids #{inspect(remaining_trip_ids)} response=#{inspect(response)}"
+            )
+
+            {remaining_trip_ids, acc_schedules, acc_trips}
```
KaylaBrady: question: Could we get stuck in an infinite loop with this?

lib/mobile_app_backend/alerts/alert_util.ex
```diff
@@ -0,0 +1,160 @@
+defmodule MobileAppBackend.Alerts.AlertUtil do
+  alias MBTAV3API.Alert
+  alias MBTAV3API.Repository
+  alias MBTAV3API.Schedule
+  alias MBTAV3API.Trip
+
+  require Logger
+
+  # Fetch schedules for all trip IDs referenced in the alert's informed_entity list
+  @spec fetch_schedules_for_alert(Alert.t(), DateTime.t()) ::
```
KaylaBrady, line 10: suggestion(non-blocking): These are covered in the higher level tests, but I think it would be worth adding a couple unit tests for these supporting functions

KaylaBrady (approved):

